PR #9, opened 2017-08-11
Implement unweighted cluster robust standard errors

lukesonnet:
This PR makes the following changes:

- Adds option to pass cluster_variable_name
- Adds either 'BM' or 'Stata' clustered errors
- Adds a `ci = T` flag that people can turn off to avoid slow confidence interval generation with 'BM' standard errors
- Changes default to return all coefficients
- Only calculates the BM DoF correction for coefficients the user asks for
- Added tests for the above

This passes the R CMD check on my machine, and it passes on AppVeyor and Travis (linux builds, mac pending). The results match the output from BMlmSE, for both the BM ses and the Stata SEs.


graemeblair:
Hi Luke, thanks, this is really exciting! Great work.

There seems to be a NOTE in appveyor, could you take a look and see how to solve it? I suspect this is a NOTE CRAN folks would be concerned about:

```* checking compiled code ... NOTE
Warning in read_symbols_from_dll(so, rarch) :
  this requires 'objdump.exe' to be on the PATH
Warning in read_symbols_from_dll(so, rarch) :
  this requires 'objdump.exe' to be on the PATH
Warning in read_symbols_from_dll(so, rarch) :
  this requires 'objdump.exe' to be on the PATH
File 'estimatr/libs/i386/estimatr.dll':
  Found no calls to: 'R_registerRoutines', 'R_useDynamicSymbols'
It is good practice to register native routines and to disable symbol
search.
See 'Writing portable packages' in the 'Writing R Extensions' manual.```

lukesonnet:
This note seems to be the result of my compilation flags which allow me to
profile on Windows. Not sure why that's made it into the package though.
I'll investigate.

Luke Sonnet
PhD Student, Political Science
UCLA

On Aug 11, 2017 19:05, "Graeme Blair" <notifications@github.com> wrote:

Hi Luke, thanks, this is really exciting! Great work.

There seems to be a NOTE in appveyor, could you take a look and see how to
solve it? I suspect this is a NOTE CRAN folks would be concerned about:

Warning in read_symbols_from_dll(so, rarch) :
  this requires 'objdump.exe' to be on the PATH
Warning in read_symbols_from_dll(so, rarch) :
  this requires 'objdump.exe' to be on the PATH
Warning in read_symbols_from_dll(so, rarch) :
  this requires 'objdump.exe' to be on the PATH
File 'estimatr/libs/i386/estimatr.dll':
  Found no calls to: 'R_registerRoutines', 'R_useDynamicSymbols'
It is good practice to register native routines and to disable symbol
search.
See 'Writing portable packages' in the 'Writing R Extensions' manual.```

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/DeclareDesign/estimatr/pull/9#issuecomment-321853287>,
or mute the thread

lukesonnet:
This seems to be a new NOTE that has been added to R to make sure methods are properly registered. The creator of Rcpp's reaction to this error was "eww" so, there's that.

 I've implemented the changes suggested [here](https://github.com/RcppCore/Rcpp/issues/651) and [here](https://stackoverflow.com/questions/42313373/r-cmd-check-note-found-no-calls-to-r-registerroutines-r-usedynamicsymbols). The solution seems straightforward enough in terms of registering these methods but still failing on Appveyor. I'll keep trying to sort out why its throwing us a note, and only for windows 32 bit architecture.

lukesonnet:
Okay, I now believe that NOTE is an AppVeyor problem. I think maybe it just doesn't have ObjDump.exe available to it or something.

I cannot recreate this note on my Windows machine, and when I submit the package to [win-builder](https://win-builder.r-project.org/) for development, it also checks fine. See the output https://win-builder.r-project.org/U9MH4HaO49Sg/ and specifically the log [here](https://win-builder.r-project.org/U9MH4HaO49Sg/00check.log) (the WARNING in that log is because of the license and the NOTE in that log is because I had to add myself as a maintainer to get an email when the build was done, that won't happen when I reset the DESCRIPTION file). Note the win-builder files will only be available for 71 hours from this comment.

Issue #3, opened 2017-08-06
Implement weighting with cluster-robust SEs


lukesonnet:
Check the R code here: https://github.com/jepusto/clubSandwich for how to implement the same small-sample correction for clustered errors with WLS.

graemeblair:
Cool. @acoppock thoughts?

acoppock:
Winston lin seyz:

I like the paper (Pustejovsky & Tipton) behind clubSandwich, but haven't tried the package. Seems worth a look …

Can we set up a Skype call to discuss pls?

Alex

graemeblair:
Yes, pls set up. I'm on European time until Saturday.

lukesonnet:
I'm in Pakistan until Friday then flying to California. Happy to meet most
Tu-Th mornings Eastern time.

Luke Sonnet
PhD Student, Political Science
UCLA

lukesonnet:
Ok, I've implemented the `clubSandwich` estimator. Our package can produce the same SEs and degrees of freedom as the `clubSandwich` package.

It is a beast though, tons of calculations and matrices held in storage. It deserves some careful time reviewing the code for inefficiencies.

Next steps:
- Working on streamlining the code for this estimator
- Removing the BM estimator altogether (results are the same if `weights = 1`; will be removed if speed is better for new estimator)
- Documenting this estimator

lukesonnet:
Note: still fails with a set of dummy variables capturing each of the clusters, the reason to implement clubSandwich (besides weighting). That's now top priority to fix.

nfultz:
overall looks good so far. I don't really have a sense of how large of objects you are operating on though. 


 also I'd recommend keeping all the debugging output, but wrapping it in an appropriate debug macro, that way it's easy to compile it back in when you want it.

graemeblair:
Thanks, Luke.

Could you say more about the debugging, Neal?

lukesonnet:
OK, the new estimator is faster as is than BM and provides equivalent results, so I'm cutting out the old implementation of BM altogether.

nfultz:
Sure, here is the debugging macro I've used:
https://github.com/nfultz/grpc/blob/master/src/common.h

Basically, you uncomment one line, recompile, and the logging is on, otherwise the macro evaluates to a blank. I stole the idiom from the `RProtobuf` package.

lukesonnet:
[...]

PR #17, opened 2017-08-20
Difference in Means Changes

lukesonnet:
Much of this is the same as the comment I left in the issue

Implemented design types:

- cluster randomized designs, for SEs used GG eq3.23
- block cluster randomized designs, for SEs used GG eq3.23 within blocks, aggregated across blocks using the usual footnote 17 formula
- matched pair unit randomized, for SEs used GG eq3.16. This is triggered simply by passing blocks that are all size 2.
- matched pair cluster randomized, for SEs used Imai-King-Nall eq. 6 with wtilde = N. If I understand it correctly, for the SATE this simplifies to m/ ((m-1)^2 * N) * sum( (N_k  (ATE_k) - N (ATE) / m)^2 ) where m is the number of matched pairs, k indexes the matched pair of clusters.

Implemented tests and errors:

- basic tests for all of the above are implemented
- errors if the same cluster is found in two blocks
- errors if treatment status differs with cluster, or is constant within block
- errors if block is size 1 unit or size 1 cluster
- errors if some blocks are size 2 and other blocks are larger
- Test that demonstrates that if cluster size is 1, then matched pair cluster design (Imai-King-Nall estimator) produces the same output of regular matched pair design (GG eq 3.16).
- Defaults to list-wise deletion for missingness on outcome or treatment, no warning
- Defaults to list-wise deletion for missingness on block, cluster, or weights variable, with warning
- Tests to make sure list-wise deletion is occurring correctly

A few questions that remain

- **Listwise deletion:** Although we allow silent list-wise deletion in `lm_robust`, should we allow it here? We might want to require cleaner data for this estimator given that the design components (cluster, blocked) can be quite particular. However, if we change how we warn/error/listwise delete we will not be consistent with `lm_robust`.
- **CIs and degrees of freedom:** Gerber and Green mention that CIs need to be increased for cluster randomized designs  by sqrt((k - 1)/(k - 2)) where k is the number of clusters. I assume this means I should add sqrt((k - 1)/(k - 2)) / 2 to each side of the interval for clustered designs. Is this correct? Is there some standard adjustments to degrees of freedom or CIs for clustered designs that you would like me to implement? More broadly, are there any other degrees of freedom adjustments, CI adjustments that should be applied to any of the blocked designs?
- **Weighting:** Weighting only works with designs with complete randomization (no cluster or blocking). Should I make this error explicit or are there some standard weighted estimators that you would like to implement for cases beyond complete randomization.
- **Testing/benchmarking:** Lastly, are there any other packages that implement a variety of DiM estimators that you would like me to benchmark this against/test for equality with?


graemeblair:
(1) Listwise deletion: let's go with consistent, unless there are specific problems we can error for related to D-I-M.
(2) CIs and degrees of freedom: can you look at `clubSandwich` and the paper associated with it for the CI adjustments for clustered designs? Happy to comment further after you have a look.
(3) Weighting: ditto for weights from that package, per your earlier comment in Issue https://github.com/DeclareDesign/estimatr/issues/3
(4) Testing/benchmarking: for a 10-unit experiment, can you show in a test that various designs (as many as you can think of) are unbiased for the true SATE? You could start with one or two to get us started.

lukesonnet:
Okay, I'll look at the `clubSandwich` package and paper next.

As for the unbiasedness of our estimate of the SATE, I'm worried that if we formalize this in a unit test with a small tolerance for unbiasendess for many designs, R CMD check will be too slow. Should I write them in the tests but then not have them run when tests are run by R CMD check?

graemeblair:
Good idea; yes let's write them but skip when running on CRAN with the `testthat` flag. Thanks, Luke.

lukesonnet:
Also, should I retract this PR and just issue one for both `lm_lin` and `difference_in_means`?

graemeblair:
If you think this is ready to merge, I'll do so now, and then later changes can come separately?

lukesonnet:
Yes, this is ready to merge although it doesn't include the above changes. I'm happy to submit `lm_lin `as a separate branch and then create a new branch for the `dim-tests` and the `dim-weighting` (these two branches will implement your above suggestions re: `clubSandwich` and unbiasedness.)

My only real concern with the branch as proposed is that the CIs might not all be the best they can be, but that will take more time to suss out.

graemeblair:
Sounds good. Let's merge now and then you can open up these various other branches. We can just wait to release on CRAN until we're sure about benchmarking/testing.

PR #21, opened 2017-08-23
Lin 2013 estimator

lukesonnet:
Implements the Lin 2013 covariate adjustment by:

1. Checking that treatment in the main formula (`formula = y ~ z`) is binary, or at least a factor with two levels.
2. Getting a matrix of all of the covariates specified like so: `covariates = ~ x1 + x2`
3. Centering that matrix by column means, naming the new columns `x1_bar, x2_bar,...` and so on
4. Column binding this data to the original data
5. Creating a new formula that interacts the treatment with the now centered covariates, like so: `y ~ z + z * x1_bar + z * x2_bar`
6. blatantly passes this new data.frame, new formula, and any other arguments (using `...` to `lm_robust`)

Passes checks for equality with manually specifying this in `lm_robust` for a variety of options.


graemeblair:
Thanks, Luke. Do you think it would be possible to do a test of the code against actual values from Lin's article, i.e. Table 2 and 3? Perhaps outside the package, so we don't have to include the data.

lukesonnet:
Sure, the data appears to be ~250kb, but I still think its better not to include it.

I think a good standard for this kind of thing would be to put it `tests/excluded_tests` and then add that to the .Rbuildignore. I'll test this out if that seems fine.

graemeblair:
That seems like a great solution. We could think about including the data later.

graemeblair:
Luke, thought about it a bit more and I think we should add Winston's data to the package. We'll send him a note after it's all set up to get his blessing. We can use it both to validate the functions in the tests but also in the examples users will see.

lukesonnet:
Sure, I had to get the data from the paper he is replicating, and I've
managed to replicate Table 2 (and almost Table 3, I'll write something
later today with more details).

Luke Sonnet
PhD Student, Political Science
UCLA

graemeblair:
Awesome, thanks so much. This should make our life easier.

lukesonnet:
Ok, we include the data with the package, automatically test it matches his Table 2 (Table 3 takes too long, but the code is in the test file).

I also closed Issue #22 and standardized the output across lm_robust and d_i_m.

I also fixed some bugs with formula parsing in lm_lin (works with factors in different places, intercept passed through transparently).

I think you should merge this in if you are ready. Then I'm going to start working on weighted clustered standard errors from `clubSandwich`.

graemeblair:
This looks really great. Excellent work. Merging in now. Thanks!

Issue #28, opened 2017-09-21
Transition to outputting "model fits" and building S3 methods


lukesonnet:
There should be a suite of classes named after the estimators that created them:

- horvitz_thompson
- difference_in_means
- lm_robust

Each of these should have their own s3 methods for:

- summary
- print
- coef
- vcov
- confint
- tidy
- predict (only for lm_robust? Actually I think we need a separate lm_lin class so we know to construct the centered and interacted data.)

Also Alex asked about anova, plot, and residuals. I think no on these, except maybe residuals on lm_robust?

Thoughts from alex that I didn't incorporate into the above:

We discussed this week that lm_robust should return a “model fit”. Normal model fits appear to be extremely heavy objects. lm fits store the data!

The value probably has to be a list (though we could have this be a dataframe with a vcov attribute!)
I don’t think we want to ape lm like syntax with summary(robust_fit)$coefficients.
I think that all computations on the data should be done when the model fit is created. Though I admit that this intuition is just coming from my aversion to carrying around the data, which seems computationally expensive.
> Luke: how expensive this is remains a ram vs. cpu problem, I'm not convinced it will be that expensive but I'll try and avoid it.

I’d also love it if the outputs of difference_in_means, horvitz_thompson and lm_robust were all unified somehow. Maybe they all have to be lists.
> Luke: I think they all have to be lists, but they can be similarly structured lists and `tidy()` will return that data frame we have always wanted <3

I think it would behoove us to take a peek around at other modelling functions in R (I guess those handled by stargazer is a good list?) and try to understand the variation in “model objects.” Maybe inspection of the broom package would also lead us in the right direction.

graemeblair:
To me the important thing is that all those s3 methods work, not that it exactly looks like an lm fit object. So avoiding summary(fit)$coefficients etc. seems reasonable -- you can get the coefficients from coef(). Let's do what is fast and simple s.t. the s3 methods still work. 

I agree it would be nice if all the functions outputs were similar; I think lists are ok if needed given now we're making it easy to interact with the outputs through s3 methods and especially `tidy()`. Is there any trickiness you're thinking of?

Looking at broom seems great. Happy to have a Skype discussion on this after you take a look if useful.

lukesonnet:
From Alex in #31:

> I think they [print and summary] should probably be the same as
> tidy(fit)
> for right now.
> 
> am I right in thinking that confint() and coef() methods are in the works too?

[...]

PR #32, opened 2017-10-13
VIgnette skeleton; S3 methods

lukesonnet:
Adds a skeleton of the vignette, building out citations and sources for the different options we provide. #10 

Also builds out the S3 methods to include coef, confint, summary, and print. #28


graemeblair:
Thanks Luke. Will wait to merge until the tests pass, I assume?

lukesonnet:
They pass. It's the same error looking for `obtain_permutation_matrix` from randomizr. There is also a warning about non-ascii characters but I'd rather resolve that when we push to CRAN as that'll just take one pass through all the code.

graemeblair:
Got it. I assume that you've tested it with the Github version of randomizr and it works so merging in.

lukesonnet:
It passes tests on my local machine with randomizr installed in the last week from github.

Issue #33, opened 2017-10-18
lm_robust breaks when cluster_variable_name is a character


jaspercooper:
... this is a problem, because `fabricatr` gives the ids as characters by default. 

```r 
rm(list = ls())
library(DeclareDesign)
library(dplyr)

population <-
  declare_population(
    cluster = level(N = 5, 
                    individuals_per_cluster = 5,
                    cluster_effect = rnorm(N)),
    individual = level(N = individuals_per_cluster,
                       noise = rnorm(N)))

potential_outcomes <- declare_potential_outcomes(
  formula = Y ~ Z + cluster_effect + noise)

assignment <- declare_assignment(clust_var = cluster, 
                                 m = 2)

estimator_factor <- declare_estimator(formula = Y ~ Z,
                               model = lm_robust,
                               cluster_variable_name = as.factor(cluster))

# Fine:
population() %>% potential_outcomes() %>% assignment() %>% reveal_outcomes() %>%  estimator_factor()

estimator_integer <- declare_estimator(formula = Y ~ Z,
                               model = lm_robust,
                               cluster_variable_name = as.integer(cluster))

# Fine:
population() %>% potential_outcomes() %>% assignment() %>% reveal_outcomes() %>%  estimator_integer()

estimator_character <- declare_estimator(formula = Y ~ Z,
                               model = lm_robust,
                               cluster_variable_name = cluster)

# Regression machine broke:
population() %>% potential_outcomes() %>% assignment() %>% reveal_outcomes() %>%  estimator_character()
```

lukesonnet:
Thanks Jasper. I thought I had resolved this but I had not. It should all work now.

Right now it takes integer, numeric, or factor variables and works quickly with them because they are all passed as NumericVector to c++.

Because c++ is strongly typed, I have to cast character clusters as factors (which can be a slowdown for long datasets).

I think there are two solutions:
1. Warn when the cluster is a character that it's faster to pass a factor, but go ahead with the conversion
2. Build a switch in the C++ that essentially will duplicate all of the cluster code (non-trivial) and add arguments for both types of cluster, string and numeric.

acoppock:
I think the right procedure here is to

cast characters as factors but do not warn.  Add to the documentation (in a couple of places!) that it’s faster if you pass factors.

THANKS

lukesonnet:
ok then that is code wise and it will be mentioned in the documentation.

PR #34, opened 2017-10-18
Refactor: clean_model_data and na.omit to track why obs are omitted

nfultz:
@lukesonnet This is the refactor I had mentioned in #33 - instead of manually removing NA in extra components (weight, cluster, ...) one at a time, we can have `model.frame` do it. I also tried to limit the NSE to one frame of the call stack at a time, so this ended up being a bit larger diff than I usually like.

All the package tests still pass and I've verified there isn't a performance hit, but I'd appreciate any feedback if there's something I overlooked.

### DIFF
--- a/R/clean_model_data.R
+++ b/R/clean_model_data.R
@@ -18,87 +18,64 @@ find_warn_missing <- function(x, type) {
 # Internal method to process data
 clean_model_data <- function(formula,
                              data,
+                             subset,
                              weights,
-                             condition_call,
-                             cluster_variable_name) {
+                             cluster) {
 
-  if (!is.null(condition_call)) {
-    r <- eval(condition_call, data)
-    data <- data[r,]
+  if(!missing(weights) && !missing(cluster) ){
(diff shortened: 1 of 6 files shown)


lukesonnet:
This looks great! I don't think you've overlooked anything except for one detail; I don't think you include in this PR the `na.omit_detailed.data.frame` method.

Also, thanks for showing me eval.parent +1.

nfultz:
Thanks, good catch 😓

lukesonnet:
Well this passes the tests that are set up to check that it works with missingness and throws appropriate warnings so it looks good. I'm going to go ahead and merge it in.

lukesonnet:
@nfultz, I'm now getting this error: `could not find function "clean_model_data"` in my luke-clubsand branch. I'm not sure what would cause this to happen as I'm still figuring out NSE, but I'm not getting this error on my machine when I install from the master branch.

I can't see anything in the diff that would cause this.

PR #35, opened 2017-10-19
Add test for na.omit_detailed, fix names bug.

nfultz:
@lukesonnet basic unit test for na.omit_detailed, and caught a bug around row names.


lukesonnet:
@nfultz Have you seen the error I'm getting with finding `clean_model_data`?

It's as simple as running the following in a clean instance of R
```r
library(estimatr)
dat <- data.frame(y = rnorm(10), x = rnorm(10))
lm_robust(y~x, data = dat)
```
Then I get:
```r
Error in clean_model_data(formula = y ~ x, data = dat, subset = , cluster = ,  : 
  could not find function "clean_model_data"
```

But as soon as I run `devtools::test()` on the directory with the package this error goes away. Can you replicate this? What is happening to the environments that is causing this error? I can't figure it out and it's keeping me from profiling the cpp code.

nfultz:
I'm getting clean builds off master, let me try your branch.

lukesonnet:
It's happening to me on the master branch as well. It's happening whether I run the above script in Rstudio or an instance of R in the terminal.

nfultz:
nm, you are right. So I think this is a NAMESPACE issue, the easiest way to
fix is to export clean_model_data, alternatively, we can change it's
signature to pass through an env.

nfultz:
If I had to guess, the reason that this passed the unit tests is because
the child environments that testthat uses are descended from the package
environment, and have access to internal / non-exported objects.

lukesonnet:
Thanks, that sounds reasonable. I was unsure about it being a NAMESPACE issue because it would work after running tests, meaning it must actually export all of the objects within the package to the global environment that I'm calling `devtools::test()` from.

Want to add the export to this PR and then I'll merge it in?

nfultz:
Sounds good for now, but I'll spend a little time on a better/longer term
solution later today.

PR #36, opened 2017-10-19
This PR threads through the source frame for NSE in clean_model_data …

nfultz:
…- instead of evaluating clean_model_data in the parent environment, which is potentially outside the package, we thread through a reference to the parent environment, then use it only for building the model.frame.

substitute is slightly annoying in this case, I had to explicitly make a `where` variable because `eval(substitute(clean_model_data(..., where=parent.frame()` would evaluate the parent.frame lazily.

@lukesonnet Do you have opinions on this one way or the other?


lukesonnet:
Ehhhhh, it's all kind of messy but I prefer this for sure.

PR #37, opened 2017-10-19
Adds the CR2 estimator proposed by authors of clubSandwich

lukesonnet:
This adds the SE estimator that can handle dummy variables that capture the clusters in the data. It also allows weighting with clustered standard errors.

I have benchmarked it agains the `clubSandwich` package and it returns the same results (and faster, presumably due to the cpp and minimizing the duplication necessary in their workflow). I can't really add these benchmark tests to the package without suggesting `clubSandwich`, but I'm happy to do so if we think it's necessary.

It also returns the same of our old implementation of the CR2 standard errors and matches the output of `BMlmSE.R` from the Green lab SOP when there is no weighting.


src/lm_robust_helper.cpp
```diff
@@ -146,24 +146,24 @@ List lm_robust_helper(const arma::vec & y,
...
-        // Cube stores the n by k G_s matrix for each cluster
-        // used for the BM dof corrction
+        // used for the dof corrction
         arma::cube H1s(k, k, J);
         arma::cube H2s(k, k, J);
         arma::cube H3s(k, k, J);
         arma::mat P_diags(k, J);
 
-        ei = y / sqrt(weights) - (X.each_col() / sqrt(weights)) * beta_hat;
+        arma::mat M_U_ct = arma::trans(arma::chol(XtX_inv));
```
nfultz: chol has a `layout` argument, instead of doing a transpose. Not sure if arma is smart enough to optimize this away.

src/lm_robust_helper.cpp
```diff
@@ -195,27 +193,25 @@ List lm_robust_helper(const arma::vec & y,
...
+          ) * X.rows(cluster_ids) ;
 
-          //arma::mat A = D * arma::sqrtmat_sympd(arma::pinv(
-          //  D * (I_min_H) * D * D
-          //)) * D * X.rows(cluster_ids) ;
+          if (ci) {
+            arma::mat ME = (XtX_inv / weight_mean) * arma::trans(At_WX);
 
-          arma::mat ME = (XtX_inv / weight_mean) * arma::trans(At_WX);
+            P_diags.col(clusternum) = arma::sum(arma::pow(ME, 2), 1);
```
nfultz: arma has a norm function for calculating two norms / frobenius norms.

src/lm_robust_helper.cpp
```diff
@@ -146,24 +146,24 @@ List lm_robust_helper(const arma::vec & y,
 
       if (type == "CR2") {
 
-        // if weights
+        // Instead of having X and Xt be weighted by sqrt(W), this has them weighted by W
         Xt = Xt.each_row() % arma::sqrt(arma::trans(weights));
+        X = X.each_col() % arma::sqrt(weights);
```
nfultz: If we never use weights not sqrt'd, we might as well pass in sqrt(w) from R

src/lm_robust_helper.cpp
lukesonnet: Thanks for that, but this summing by row. Can it handle that?

src/lm_robust_helper.cpp
lukesonnet: Great, thanks.

src/lm_robust_helper.cpp
lukesonnet: Great, there was also a duplication of effort in R. Thanks.

PR #38, opened 2017-10-19
Worst demo ever, and use it to test that things work in Global environment

nfultz:
@lukesonnet this is a very basic test for making sure things don't break when calling from the Global environment.  

Also feel free to copy some of your testing scripts to the demo/ folder


lukesonnet:
How do demos normally relate to tests and vignettes? Is the purpose simply to have manually run tests that check the things work in the global environment?

nfultz:
My current understanding is:

tests - verifies things aren't broken - implemented either via testthat or
traditional tests in the test/ directory - run manually, by R CMD check, by
CRAN, or by CI before build/shipping. End users typically don't run these.

demos - extended things a user might want to do, possibly interactive -
basically just scripts that get sourced as if a user had sourced it
themselve - run only when a user types demo()

vignettes - documentation mixed with code - usually supplemental to man
pages and / or R-Journal or JSS articles - run at build time, static
html/pdf file displayed when user types vignette(), also mirrored on CRAN
IIRC.

examples - these live inside the function .Rd pages - for us they can come
from a roxygen @examples block - run when the package is built, and when
user runs example(foo)

so to a large degree, there's a four-circle venn-diagram of where different
snippets of code could live .

For the particular NSE bug I caused, (where because of parent environments
a  testthat passed where a demo would have failed because demos run in the
user environment and testthat runs in a child of the package), I think
there's value in testing demos.

PR #41, opened 2017-10-21
Update README.md

nfultz:
Currently the drat repo is serving pretty old versions of packages. This gives people a clue on how to get updated packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,10 @@ Technical papers and textbooks demand complex estimation strategies that are oft
 To install the latest development release of **estimatr**, please ensure that you are running version 3.3 or later of R and run the following code:
 
 ```
+# Development branch
+devtools::install_github("DeclareDesign/estimator")
+
+# Last release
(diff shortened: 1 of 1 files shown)


graemeblair:
Actually, @aaronrudkin could you work on fixing the drat repository instead of switching to this? We made the switch because estimatr requires installing devtools and it was requiring lots of user support to help people. This provides binaries instead.

nfultz:
I'll take a look at the drat repo today, can you provide a bit of
background on what's wrong with it?

graemeblair:
Sure sounds good, thanks @nfultz . I'm not sure what's wrong, but @aaronrudkin set up a great system using Travis and Appveyor that automatically pushes the new binary from OS X / Windows onto the declaredesign.github.io repo. So the place to start is the travis and appyeyor yml's. I can add you to the various travis/appveyor accounts as needed if you drop by my office.

lukesonnet:
Maybe because the checks aren't passing they aren't pushing the new binaries?

PR #45, opened 2017-10-26
Lukesonnet/fix cmd

lukesonnet:
Fixes error in clean model data with blocking


lukesonnet:
Sorry just did this myself as it's a simple fix to a bad error.

Issue #59, opened 2017-11-21
Ad DV name to tidy output


acoppock:
It'd be awesome if there were a column in the output called `dv` that reported the name of the dependent variable.

hashtag feature request.

acoppock:
I was thinking of adding a `dv` column to the tidy output of `lm_robust` and other estimators. right now the output looks like this:

```
 coefficient_name       est        se           p   ci_lower  ci_upper  df
1      (Intercept) 0.2222222 0.1008317 0.028978271 0.02307035 0.4213741 158
2   swap_treatment 0.2918623 0.1092642 0.008350522 0.07605537 0.5076692 158
```

nfultz:
:+1: - keep getting confused on which ticket is which repo, I mostly reply to these in gmail, if the text doesn't say it's pretty hard to tell.

lukesonnet:
Do you really want this? Can I add it as an attribute or do you want it to be more visible.

acoppock:
I think it should be in the data.frame, not an attribute.

It’s something that I’m *constantly* adding when preparing model fits for either meta analysis or plotting.  It’s also a good thing to remind users in the output what they were regressing on what!

Thanks!

lukesonnet:
can I call it `outcome` rather than `dv`? Your call.

lukesonnet:
Added in #62

PR #62, opened 2017-11-30
Add S3 predict method

lukesonnet:
This PR adds a predict method that largely mimics `predict.lm`. It works with both `lm_robust` and `lm_lin`. This function differs from `predict.lm` in the following ways:

- You cannot specify your own residual variance; i'm not sure when one would want to do this anyways. To utilize this, we would need to keep the original design matrix.
- You cannot specify a `scale` parameter. To utilize this, we would need to keep the original design matrix.
- It will provide different SEs and intervals if you use robust standard errors because of how we calculate the standard error of the prediction. What we do mimics stata's functionality which can be seen describing `stdp` on page 11 of the predict manual [here](https://www.stata.com/manuals13/rpredict.pdf).  Nonetheless, with classical SEs, ses of the fit and prediction and confidence intervals are always the same.

And lastly, you have to specify `newdata`, you can't just use the original model.frame. I think this is fine, because they should have the data anyways. Furthermore, this way, we have still avoided adding *anything* with length `n` to the output of `lm_robust`.

This is documented with examples, details, and a test file is also created.


R/predict.R
```diff
@@ -0,0 +1,152 @@
...
+
+  df_resid <- object$n - object$rank
+  interval <- match.arg(interval)
+
+  if (se.fit || interval != "none") {
+
+    ret <- list()
+
+    var_fit <-
+      apply(X, 1, function(x) tcrossprod(crossprod(x, object$vcov), x))
```
nfultz: for ```x'Sx``` I usually group it the other way - the parens here are a little confusing to me, it looks like you wrote ```x'Sx'``` which seems odd.

R/predict.R
```diff
@@ -0,0 +1,152 @@
...
+        # Get weights
+        if (missing(weights)) {
+          weights <- 1
+        } else {
+          weights <- eval(substitute(weights), newdata)
+        }
+
+        pred_var <-
+          if (is.null(pred.var)) {
+            object$res_var / weights
```
nfultz: no need for ternary here, `if(is.null(pred.var)) { pred.var <- ... } `

R/predict.R
```diff
@@ -0,0 +1,152 @@
...
+      if (interval == "prediction") {
+
+        if (missing(weights) &&
+            object$weighted &&
+            is.null(pred.var)) {
+          warning("Assuming constant prediction variance even though model fit is weighted\\n")
+        }
+
+        # Get weights
+        if (missing(weights)) {
```
nfultz: You can combine the two `if(missing(weights))` statements

lukesonnet:
I've solved #59 in here as well.

All methods now return `outcome` in the list of output which is a string with the name of the output variable (note that if you pass `as.factor(x)` it will just say `x` as the outcome).

`tidy` now cbinds this to the end of the dataframe, and so `print` will print it. However, `summary` drops it because it makes the summary matrix a character matrix which is no good.

I could add it as second element in the `summary` list but I don't think that's necessary.

R/predict.R
```diff
@@ -0,0 +1,152 @@
...
+            pred.var
+          }
+
+        hwid <- tval * sqrt(var_fit + pred_var)
+      } else if (interval == "confidence") {
+        hwid <- tval * sqrt(var_fit)
+      }
+
+      predictor <- cbind(predictor,
+                         predictor + hwid %o% c(1, -1))
```
nfultz: just make three columns explicitly, ```data.frame(fit=predictor, lwr=predictor - hwid, upr=predictor + hwid)```

nfultz: or a matrix if that's your intent.

NAMESPACE
```diff
@@ -30,12 +31,17 @@ export(lm_robust_fit)
 export(permutations_to_condition_pr_mat)
 export(tidy)
 importFrom(Rcpp,evalCpp)
+importFrom(stats,.checkMFClasses)
```
nfultz: Be careful about import dotted things, that may trip a cran warning.

R/horvitz_thompson.R
```diff
@@ -332,9 +332,9 @@ horvitz_thompson <-
     # Build and return output
     #-----
 
-    return_list$alpha <- alpha
-
-    return_list$coefficient_name <- all.vars(formula[[3]])
+    return_list[["alpha"]] <- alpha
+    return_list[["coefficient_name"]] <- all.vars(formula[[3]])
+    return_list[["outcome"]] <- all.vars(formula[[2]])
```
nfultz: duplicated code as difference in means? maybe write a fn.

R/lm_lin.R
```diff
@@ -65,7 +64,7 @@ lm_lin <- function(formula,
   cluster <- model_data$cluster
 
   # If Z is a factor, can't use variable name
-  # So get first column non intercept column
+  # So get first non-intercept column (always will be treatment)
```
nfultz: This is a reasonable assumption, make sure it's documented in the help page though - sometimes people get their factor levels out of order..

R/lm_lin.R
```diff
@@ -113,5 +114,12 @@ lm_lin <- function(formula,
       return_vcov = return_vcov
     )
 
+  return_list[["contrasts"]] <- attr(model_data$design_matrix, "contrasts")
+  return_list[["terms"]] <- model_data$terms
+  return_list[["weights"]] <- model_data$weights
+  return_list[["outcome"]] <- all.vars(formula[[2]])
+  return_list[["scaled_center"]] <- attr(demeaned_covars, "scaled:center")
+  names(return_list[["scaled_center"]]) <- original_covar_names
```
nfultz: consider setNames instead for clarity - I try to avoid setter functions when possible.

R/lm_robust_fit.R
```diff
@@ -138,15 +140,27 @@ lm_robust_fit <- function(y,
...
     )
 
   if (return_vcov & se_type != 'none') {
+    #return_list$residuals <- fit$residuals
     return_list$vcov <- fit$Vcov_hat
     dimnames(return_list$vcov) <- list(return_list$coefficient_name,
                                        return_list$coefficient_name)
   }
 
+  if (!is.null(weights)) {
```
nfultz:
```r
return_list$weighted <- !is.null(weights)
if(return_list$weighted) {
  return_list$res_var <- ...
}
```

NAMESPACE
lukesonnet: It doesn't warn with `R CMD check`, and not including it leaves a warning (or a note, I forget which). So I guess we should leave it for now?

R/lm_lin.R
lukesonnet: We only accept something that has one variable on the rhs, but I guess you just made it clear that if that was a factor it could create multiple columns in the model matrix. Not sure how to handle this but I'll think about it.

R/predict.R
lukesonnet: It's xSx', where x is a row-vector.

The reason I do `crossprod(x, object$vcov)` is that crossprod will take `x` to be a column vector when I want it to be a row vector.

R/predict.R
lukesonnet: Leaving as a matrix to match `predict.lm`.

lukesonnet:
Thanks for the review Neal. Appreciate it.

I've incorporated his comments and Neal you should feel free to merge this in.

nfultz:
Easy failure to fix?
```
1. Failure: Test LM Lin (@test-lm-lin.R#39) ------------------------------------
  error$message does not match "no more than two levels".
  Actual value: "Treatment variable must be binary for the Lin estimator."
```

R/predict.R
nfultz: in R vectors are vectors, you usually don't need to worry about row/column vectors when writing code, but going backwards from code to math is difficult. EG:

```
> S <- matrix(1:9,3,3)
> x <- 1:3
> tcrossprod(crossprod(x, S), x)
     [,1]
[1,]  228
> x %*% S %*% x
     [,1]
[1,]  228
> x %*% crossprod(S, x) 
     [,1]
[1,]  228
```

If the intent is `xSx'` (with one transpose mark in the equation at the very end) I would suggest - 

```
>  tcrossprod( x %*% S, x)
     [,1]
[1,]  228
```

bc `x %*% tcrossprod(S,x)` throws an error.



as it's the most literal translation.

PR #65, opened 2017-12-13
Migrate to Eigen; allow rank-deficient X matrices

lukesonnet:
This is a fairly major rewrite of the `lm_robust_fit.R` and `lm_robust_helper.cpp` files. This essentially migrates all of our LS solver and standard errors to use `RcppEigen` in order to take advantage of the rank-revealing QR decomposition it has. Major changes:

1. Allows for solutions involving rank-deficient X matrices, similar to R's `lm` where we NA for some of the singular matrices, and print a message letting the user know we've done this.
2. Allows users to specify `trychol = TRUE` if they want to try a cholesky decomposition before the QR. I believe that this LLt decomposition does not always fail when X is rank-deficient and thus we cannot rely on it to identify rank-deficiencies. However, the LLt decomposition is much faster than the RRQR decomposition, so power users who are confident they have full-rank X can specify `trychol = TRUE` and enjoy speed gains. We should discuss this further in our weekly meeting.
3. For `horvitz_thompson()`, this fixes the `get_condition_pr_mat` function for complete randomizations; the formula was incorrect before.
4. Completely replaces `RcppArmadillo` dependency with `RcppEigen`.

This also adds tests to make sure that the S3 methods work with the rank-deficiency, that we match `RcppEigen::fastLm()` in rank-deficient cases (because the QR in Eigen and the QR in lm don't always agree when there is a rank-deficiency!). Also fixed a few small bugs here and there.


R/lm_robust.R
```diff
@@ -25,7 +25,9 @@ lm_robust <- function(formula,
                       ci = TRUE,
                       alpha = .05,
                       coefficient_name = NULL,
-                      return_vcov = TRUE) {
+                      return_vcov = TRUE,
+                      ei = TRUE,
+                      trychol = FALSE) {
```
lukesonnet: FYI, ei means use Eigen and `lm_speed.cpp` versus the old algorithm in `lm_robust_helper.cpp` (i.e. armadillo `solve(Xt*X)`).

`trychol` lets `lm_sped.cpp` know whether it should try the cholesky decomposition first.

R/lm_robust_fit.R
```diff
@@ -65,58 +67,112 @@ lm_robust_fit <- function(y,
...
+      cluster <- cluster[cl_ord]
+      J <- length(unique(cluster))
+      if (!is.null(weights)) {
+        weights <- weights[cl_ord]
+      }
+
+    } else {
+      J <- 1
+    }
+
```
lukesonnet: Here is where I reorder everything so that I only have to pass through the data once in RcppEigen.

src/lm_speed.cpp
```diff
@@ -0,0 +1,502 @@
...
+  int j = 0;
+  for (int i = 0; i < full.rows(); i++)
+  {
+    if (ind(i) == 1) {
+      target.row(j) = full.row(i);
+      j++;
+    }
+  }
+  return target;
+}
```
lukesonnet: These two functions aren't used right now, but they were how I was subsetting in Eigen before.

src/lm_speed.cpp
```diff
@@ -0,0 +1,502 @@
...
+  bool full_rank = true;
+
+  // Many of the lines below are inspired by the fastLm function from RcppEigen
+  // https://cran.r-project.org/web/packages/RcppEigen/vignettes/RcppEigen-Introduction.pdf
+  if (chol) {
+    const Eigen::LLT<Eigen::MatrixXd> llt(Xfullt*Xfull);
+    beta_out = llt.solve(Xfull.adjoint() * y);
+    R_inv = llt.matrixL().solve(Eigen::MatrixXd::Identity(p, p));
+    XtX_inv = R_inv.transpose() * R_inv;
+  } else {
```
lukesonnet: This block is basically just "only run and try cholesky" and I don't think we will need this going forward.

src/lm_speed.cpp
```diff
@@ -0,0 +1,502 @@
...
+    try {
+      if (trychol) {
+        const Eigen::LLT<Eigen::MatrixXd> llt(Xfullt*Xfull);
+
+        // Catch case where Xfull is rank-deficient
+        if(llt.info() == Eigen::NumericalIssue) {
+          full_rank = false;
+          throw std::runtime_error("Possibly non semi-positive definite matrix!");
```
lukesonnet: Force an exception if there is a "NumericalIssue" with the eigen decomposition.

src/lm_speed.cpp
```diff
@@ -0,0 +1,502 @@
...
+  } else {
+    beta_hat = beta_out;
+    X = Xfull;
+    if (weighted) {
+      Xoriginal = Xoriginalfull;
+    } else {
+      Xoriginal = Xfull;
+    }
+  }
+
```
lukesonnet: Here is where I construct all of my matrices for the work below. Note that I am just dropping the columns of X that have been dropped by the QR, and removing the NAs from beta.

src/lm_speed.cpp
```diff
@@ -0,0 +1,502 @@
...
+               Eigen::MatrixXd H1t = H1s.row(k);
+               Eigen::MatrixXd H2t = H2s.row(k);
+               Eigen::MatrixXd H3t = H3s.row(k);
+               // Rcout << H1t << std::endl<< std::endl;
+
+               H1t.resize(r, J);
+               H2t.resize(r, J);
+               H3t.resize(r, J);
+
```
lukesonnet: I had used cubes before for the H1, H2, and H3 matrices but now I have to rely on just stacking each slice (3rd dimension) into more columns and then resizing.

src/lm_speed.cpp
```diff
@@ -0,0 +1,502 @@
...
+            len++;
+            continue;
+          }
+          //Eigen::VectorXi cluster_ids = (clusters == levels[i]).cast<int>();
+
+          //Eigen::VectorXd ei_clus = extract_vec(ei, cluster_ids);
+          //Eigen::MatrixXd X_clus = extract_mat_rows(X, cluster_ids);
+
+        }
+        // Rcout << "XteetX: " << XteetX << std::endl;
```
lukesonnet: This is a simple way to see the algorithm I use for looping over clusters.

src/lm_speed.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+               Eigen::MatrixXd uf = H1t.transpose() * H2t;
+               Eigen::MatrixXd P_row = P_diags.row(k).asDiagonal();
+               Eigen::MatrixXd P_array = (H3t.transpose()*H3t - uf - uf.transpose()) + P_row;
+               // Rcout << "P_array: " << P_array << std::endl;
+
+               // Rcout << std::pow(P_array.trace(), 2) << " and " << P_array.array().pow(2).sum() << std::endl;
+               dof(k) = std::pow(P_array.trace(), 2) / P_array.array().pow(2).sum();
+             }
+             k++;
```
nfultz: Is k++ in right spot - eg increment even if whichcov[j] is false?

R/lm_robust_fit.R
```diff
@@ -147,11 +203,12 @@ lm_robust_fit <- function(y,
...
     )
 
+
   if (return_vcov & se_type != 'none') {
     #return_list$residuals <- fit$residuals
     return_list$vcov <- fit$Vcov_hat
-    dimnames(return_list$vcov) <- list(return_list$coefficient_name,
-                                       return_list$coefficient_name)
+    dimnames(return_list$vcov) <- list(return_list$coefficient_name[est_exists],
+                                       return_list$coefficient_name[est_exists])
```
nfultz: Usually use `&&` on if statements.

lukesonnet:
Made most of @nfultz's recommended changes. I'm going to merge this at end-of-day unless someone says otherwise! We have some things to discuss tomorrow, though, which I will highlight in Slack.

graemeblair:
It links to and suggests `RcppEigen`?

lukesonnet:
Yes, because it needs to be available for a test and LinksTo doesn't do that. I could just import it but that's unnecessary for the rest of the package to work.

PR #67, opened 2017-12-18
Update appveyor.yml

nfultz:


nfultz:
AHHH, this was only for debugging, let's back it out.

nfultz:
It will cause all your windows builds to hang for an hour.

Issue #50, opened 2017-11-04
Error in appveyor when checking PRs


lukesonnet:
There's an error that only exists in a PR, and not the branch itself or the resulting merged branch (which I suppose the pr check is supposed to be doing). Seems to be a problem with the drat repo?

See it around like 400 here: https://ci.appveyor.com/project/DeclareDesign/estimatr/build/1.0.201/job/cw717ysnrufj9pn2

I'm not sure of the nitty gritty here but I'll get to this around when prepping for CRAN submission if no one takes it on before then, or feel free to close this if this is part of a larger issue that people are aware of.

lukesonnet:
Seems like @nfultz has solved this.

Issue #51, opened 2017-11-04
Error in Travis-CI when building vignettes


lukesonnet:
Because this vignette will have a lot of math, I think it's best to build it as a PDF unless there are objections. If there are, perhaps we can discuss them in #10.

However, because it is a PDF, it is built using LaTeX and there is an error in the Mac set-up on Travis-CI (seen [here](https://travis-ci.org/DeclareDesign/estimatr/jobs/297265522)) where we basically only need to add texlive-extras to the install unless there are other suggestions.

However, we don't currently use a .travis.yml so I'm not sure where these settings exist. @graemeblair?

graemeblair:
They're in .travis.yml in `estimatr` (we do use this). Neal is working on a docker replacement for how we're using travis right now, so I think if you are good to stick with your solution for the meantime that would avoid figuring out these issues.

lukesonnet:
Sorry, still figuring out why my Mac sometimes shows hidden files and why other times it does not. I'll close this for now and we can revisit these bugs post nfultz.

aaronrudkin:
Did we ever get a solution on this? I tried to add a few short equations to the documentation of a couple functions in fabricatr and ran into a bunch of complaining from travis. They're not crucial to functionality, but I'd like it so that when I look at the code a year from now I remember what goes on in the math.

PR #71, opened 2017-12-20
Minor Edits

lukesonnet:
Changes a warning message with rank-deficient X, renames the argument `trychol`, and replaces a link in the readme to be absolute rather than relative so it works on both declaredesign.org and github.

Will merge once I check on the automated checks (and if they finally work in a PR, then I'll close issues #50 and #51).

### DIFF
--- a/R/tidy.R
+++ b/R/tidy.R
@@ -114,6 +114,7 @@ tidy_data_frame <- function(object, digits = NULL) {
 warn_singularities <- function(object) {
   if (object$rank < object$k) {
     singularities <- object$k - object$rank
-    message(sprintf('%i coefficient%s not defined because the design matrix is rank deficient\n', singularities, ifelse(singularities > 1, 's', '')))
+    plural <- ifelse(singularities > 1, "s", "")
+    message(singularities, " coefficient", plural, " not defined because the design matrix is rank deficient\n")
   }
 }
--- a/R/tidy.R
+++ b/R/tidy.R
@@ -114,6 +114,6 @@ tidy_data_frame <- function(object, digits = NULL) {
 warn_singularities <- function(object) {
   if (object$rank < object$k) {
     singularities <- object$k - object$rank
-    message(sprintf('%i coefficient%s not defined because of singularities\n', singularities, ifelse(singularities > 1, 's', '')))
+    message(sprintf('%i coefficient%s not defined because the design matrix is rank deficient\n', singularities, ifelse(singularities > 1, 's', '')))
   }
 }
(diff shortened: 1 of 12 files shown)


R/tidy.R
```diff
@@ -114,6 +114,6 @@ tidy_data_frame <- function(object, digits = NULL) {
 warn_singularities <- function(object) {
   if (object$rank < object$k) {
     singularities <- object$k - object$rank
-    message(sprintf('%i coefficient%s not defined because of singularities\n', singularities, ifelse(singularities > 1, 's', '')))
+    message(sprintf('%i coefficient%s not defined because the design matrix is rank deficient\n', singularities, ifelse(singularities > 1, 's', '')))
```
nfultz: you usually don't need/want to sprintf inside message, you can just do `message(singularities, " cofficient", plural, " blah blah...")` - this will auto generate format strings for i18n should we ever do that.

lukesonnet: Thanks, old habits die hard.

README.Rmd
```diff
@@ -58,7 +58,7 @@ lm_robust(y ~ z, data = sample_dat, clusters = my_cluster_var)
 difference_in_means(y ~ z, data = sample_dat, blocks = my_block_var)
 ```
 
-The [Getting Started](articles/estimatr-vignette.html) guide describes each estimator provided by **estimatr** and how it can be used in your analysis.
+The [Getting Started](http://estimatr.declaredesign.org/articles/estimatr-vignette.html) guide describes each estimator provided by **estimatr** and how it can be used in your analysis.
```
nfultz: I feel like the link text should be "Getting Started Guide" instead of just "Getting Started" although opinions may vary.

lukesonnet: Agree

src/lm_robust_helper.cpp
```diff
@@ -35,7 +35,7 @@ List lm_solver(Eigen::Map<Eigen::MatrixXd>& Xfull,
   // Much of the OLS solution code is inspired by or copied directly from the fastLm function from RcppEigen
   // https://cran.r-project.org/web/packages/RcppEigen/vignettes/RcppEigen-Introduction.pdf
   try {
-    if (trychol) {
+    if (try_cholesky) {
```
nfultz: Catching your own exception as below is a bit gross - should probably break this in to smaller functions at some point.

lukesonnet: I agree. I'll return this when I sweep through again.

graemeblair:
looks good to merge @lukesonnet yes?

nfultz:
merging now.

PR #73, opened 2018-01-05
Minimal working example of margins compatability

nfultz:
@lukesonnet This gets us started with using `lm_robust()` models with the `margins` package by @leeper - `marginal_effects` by default finds the training data set from `model$call`, and everything else seemed to work. 

If I understand correctly, `lm_robust` should have identical coefficients as `lm`, and should only differ in standard errors / width of prediction intervals. 

Currently,  when `clusters` are specified the cluster variable is not stripped out - however `weights` do work because there is already special logic for it in `margins:::find_terms_in_model`.

### DIFF
--- a/R/lm_robust.R
+++ b/R/lm_robust.R
@@ -59,5 +59,8 @@ lm_robust <- function(formula,
                            model_data = model_data,
                            formula = formula)
 
+
+  return_list$call <- match.call()
+
   return(return_list)
 }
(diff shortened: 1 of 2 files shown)


nfultz:
It looks like I'll need to rework a few tests, now that `call` is returned it's causing some diffs between otherwise identical models.

lukesonnet:
Sorry I never reviewed this. I'll merge it in with a new branch in a minute and get everything working as best I can.

lukesonnet:
Now merged in to my branch. So long and thanks for all the fish!

PR #75, opened 2018-01-08
Adds support for multi-valued treatments in lm_lin

lukesonnet:
This adds support for multi-valued treatments with `lm_lin` where the treatments can be passed as any kind of vector. Also adds tests.


R/lm_lin.R
```diff
@@ -61,38 +61,56 @@ lm_lin <- function(formula,
...
-    )
+    if (ncol(treatment) > 1) {
+      stop(
+        "Treatment variable must be binary or a factor for the Lin estimator."
+      )
+    } else {
+      # create dummies for non-factor treatment variable
+
+      # Drop out first group if there is an intercept
+      vals <- sort(unique(treatment))[if(has_intercept) {-1} else {TRUE}]
```
nfultz: split into two lines, `if(has_intercept) vals <- vals[-1]`

R/lm_lin.R
```diff
@@ -101,16 +119,45 @@ lm_lin <- function(formula,
...
+    cols <- (i - 1) * n_treat_cols + (1:n_treat_cols)
+    interacted_covars[, cols] <-  apply(treatment, 2, `*`, demeaned_covars[, i])
+    interacted_covars_names[cols] <- paste0(colnames(treatment), ":", covar_name)
+  }
+  colnames(interacted_covars) <- interacted_covars_names
+  #print(interacted_covars)
+
+  if (has_intercept) {
+    # Have to manually create intercept if treatment wasn't a factor
+    X <- cbind(matrix(1, nrow = n, ncol = 1, dimnames = list(NULL, c("(Intercept)"))),
```
nfultz: gratuitous c around intercept.

R/lm_lin.R
```diff
@@ -101,16 +119,45 @@ lm_lin <- function(formula,
...
+  for (i in 1:n_covars) {
+
+    covar_name <- colnames(demeaned_covars)[i]
 
-  # Interact with treatment
-  X <- cbind(design_matrix[, attr(design_matrix, "assign") <= 1, drop = F],
-             demeaned_covars,
-             interacted_covars)
+    cols <- (i - 1) * n_treat_cols + (1:n_treat_cols)
+    interacted_covars[, cols] <-  apply(treatment, 2, `*`, demeaned_covars[, i])
```
nfultz: ``` treatment * demeaned_covars[,i] ``` would be easier to read?

R/lm_lin.R
```diff
@@ -61,38 +61,56 @@ lm_lin <- function(formula,
...
+      treatment_mat <- matrix(
+        NA,
+        nrow = n,
+        ncol = n_treats,
+        dimnames = list(NULL,
+                        paste0(colnames(design_matrix)[treat_col], vals))
+      )
+
+      for (i in 1:n_treats) {
+        treatment_mat[, i] <- as.numeric(treatment == vals[i])
```
nfultz: Can vectorize the loop with outer - ```treatment_mat[] <- outer(treatment, vals, `==`) ```

R/lm_lin.R
lukesonnet: It doesn't do the correct operation if treatment has multiple columns.

R/lm_lin.R
lukesonnet: Note that treatment is a matrix with multiple columns!

R/lm_lin.R
nfultz: oy

nfultz: what do you mean? Am I missing something?
```

> X <- matrix(1:4, 2,2)
> X
     [,1] [,2]
[1,]    1    3
[2,]    2    4
> X * 1:2
     [,1] [,2]
[1,]    1    3
[2,]    4    8
> apply(X, 2, `*`, 1:2)
     [,1] [,2]
[1,]    1    3
[2,]    4    8
```

R/lm_lin.R
nfultz: wait, if `treatmat_mat[, i]` is not the same length as `as.numeric(treatment == vals[i])` that also seems bad.

nfultz: let's sleep on this and make it less idiosyncratic.

R/lm_lin.R
lukesonnet: Hmmm, maybe at some point I had it multiplying by all of demeaned_covars instead of one column at a time. Your suggestion should work.

R/lm_lin.R
```diff
@@ -82,21 +82,15 @@ lm_lin <- function(formula,
...
-        dimnames = list(NULL,
-                        paste0(colnames(design_matrix)[treat_col], vals))
-      )
+      # TODO warn if too many values? (ie. if there are as many treatments as observations, maybe we should warn)
 
-      for (i in 1:n_treats) {
-        treatment_mat[, i] <- as.numeric(treatment == vals[i])
-      }
+      # Create matrix of dummies
+      treatment_mat <- outer(drop(treatment), vals, function(x, y) as.numeric(x == y))
```
nfultz:
```
treatment_mat <- 1.0 * outer(X, Y, `==`)
```

R/lm_lin.R
```diff
@@ -134,15 +128,14 @@ lm_lin <- function(formula,
...
+    interacted_covars[, cols] <- treatment * demeaned_covars[, i]
     interacted_covars_names[cols] <- paste0(colnames(treatment), ":", covar_name)
   }
   colnames(interacted_covars) <- interacted_covars_names
-  #print(interacted_covars)
 
   if (has_intercept) {
     # Have to manually create intercept if treatment wasn't a factor
-    X <- cbind(matrix(1, nrow = n, ncol = 1, dimnames = list(NULL, c("(Intercept)"))),
+    X <- cbind(matrix(1, nrow = n, ncol = 1, dimnames = list(NULL, "(Intercept)")),
```
nfultz: Also, you can let cbind set the name on the intercept column for you -
```
> cbind(`(Intercept)`=1, sleep)
   (Intercept) extra group ID
1            1   0.7     1  1
2            1  -1.6     1  2
3            1  -0.2     1  3
```

nfultz: nm, that only works on dfs.

R/lm_lin.R
nfultz: `outer` can also set the names correctly, as in:
```
> outer(setNames(nm=LETTERS[1:3]), setNames(nm=letters[1:3]), `==`)
      a     b     c
A FALSE FALSE FALSE
B FALSE FALSE FALSE
C FALSE FALSE FALSE
```

so you could do the naming a little earlier
```
names(vals) <- paste0(colnames(design_matrix)[treat_col], vals)
outer(drop(treatment), vals)
```

and at this point, there's enough logic around vals to justify a helper function - `get_vals_from_treat_matrix()` or similiar.

lukesonnet: Is there any reason to prefer 1.0 to as.numeric?

PR #80, opened 2018-01-17
Solve bug with clustered SEs

lukesonnet:
If last cluster (numerically) was a singleton, then it was skipped in the estimation of cluster robust variance. This fixes that problem (and as an added benefit simplifies the algorithm).

This should be merged ASAP (lets let atleast AppVeyor check it), I think.

### DIFF
--- a/src/lm_robust_helper.cpp
+++ b/src/lm_robust_helper.cpp
@@ -335,8 +335,8 @@ List lm_solver(Eigen::Map<Eigen::MatrixXd>& Xfull,
        }
 
 
-       Rcout << "XtX_inv: " << std::endl << XtX_inv << std::endl;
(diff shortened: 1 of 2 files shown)


src/lm_robust_helper.cpp
```diff
@@ -243,16 +243,16 @@ List lm_solver(Eigen::Map<Eigen::MatrixXd>& Xfull,
 
        // iterate over unique cluster values
        for(int i = 1;
-           i < n;
+           i <= n;
```
nfultz: for loops can be one line.

lukesonnet: lol no idea why it's more than one line. Must have had something long previously....

lukesonnet:
Linux and Windows passed. Merging.

Issue #78, opened 2018-01-16
broom integration


acoppock:
is apparently BROKE.

MWE:

```
library(estimatr)
library(broom)

X <- rnorm(100)
Y <- rnorm(100)

fit <- lm_robust(Y ~ X)
tidy(fit) # errors
estimatr::tidy(fit)
```

lukesonnet:
There's a decent explanation of this here: https://stackoverflow.com/a/18515795/4257918

Here's the problem:

When you load `broom` after `estimatr`, it clobbers the generic `tidy` method. Therefore, you are now calling `broom::tidy` whenever you call `tidy`. The problem is that both `broom::tidy` and `estimatr::tidy` are the following function:

```
tidy <- function(object, ...) UseMethod("tidy")
```
As per the `UseMethod` help docs:
```
To support this, UseMethod and NextMethod search for methods in two places: first in the environment in which the generic function is called, and then in the registration data base for the environment in which the generic is defined (typically a namespace).
```
So this means that when you call `broom::tidy`, `UseMethod` will only search the environment in which it is called and the `package:broom` namespace, because that is where `broom::tidy` is defined.

So there are two solutions:
1. Add a broom dependency
2. Forcefully export `tidy.lm_robust` (and the other methods) to the environment in which users load `estimatr`. This means when the user calls `broom::tidy` it is content to look in the current environment, to which we now forcefully export `tidy.lm_robust`. This doesn't solve the reverse problem, where if you load our package after theirs, then we have clobbered their default `tidy` method, and we can't force them to export all of their `tidy.lm` functions.

lukesonnet:
Neal when you look at this, could you let me know what about what I said in the meeting was technically incorrect? I've just cobbled together what I could from man pages and Adv-R, but I couldn't find an easy resource to explain this to me.

nfultz:
just to be explicit, we have three use cases -

```
require(estimatr)

model2 <- lm_robust(extra~group, sleep, clusters = "ID")

environment(tidy)
tidy(model2)


#vs

require(broom)
require(estimatr)

model <- lm(extra~group, sleep)
model2 <- lm_robust(extra~group, sleep, clusters = "ID")

environment(tidy)
tidy(model)
tidy(model2)

#vs

require(estimatr)
require(broom)

model <- lm(extra~group, sleep)
model2 <- lm_robust(extra~group, sleep, clusters = "ID")
environment(tidy)
tidy(model)
tidy(model2)

```

[...]

PR #82, opened 2018-01-18
Deals with broom package

lukesonnet:
Deals with broom conflict (Issue #78) by making tidy.<estimator> available and allowing users to use broom::tidy.default if they have loaded broom.

Also adds tests for this that aren't run by `R CMD check` but do run on a local machine when `devtools::test()` is called. This is to save us from "suggesting" broom.


lukesonnet:
Linux and Windows passing; merging.

Issue #18, opened 2017-08-21
horvitz_thompson


graemeblair:
We'd like to implement `horvitz_thompson` as a second estimator, mimicking the functionality of `difference_in_means`.

lukesonnet:
I think a bit more guidance on this would be appreciated. I'm not sure how this would mimick the functionality of `difference_in_means` if it is primarily used to estimate totals. Do you think you could provide a reference that discusses the usage the way you would like it implemented?

Thank you. Also, an eventual question I think will appear later: for the estimate of the variance, should I by default assume that the weights/probabilities are known and fixed when passed by the user?

lukesonnet:
Even an example of an application would likely be all I need; this isn't an estimator I'm familiar with.

graemeblair:
It can be used as a drop-in replacement for `difference-in-means` for estimating the difference in means. I think the usage will be identical or nearly identical. The estimator for the totals can be divided by N to obtain the estimator for the mean. I think the discussion in the Imbens-Rubin book is good, but will check tomorrow when in the office.

Yes, you can assume that `Pr(T_i=1)` is fixed and known.

lukesonnet:
Ok, just checked out Imbens-Rubin. I'll implement eqn 17.8 as the estimator, but IR aren't very clear about the variance.

Looking at the estimator of the variance proposed by Aronow and Middleton [here](http://aronow.research.yale.edu/unbiased.pdf), but not sure if that's the variance (eq 35) we want to use, It seems to be the most appropriate.

Another question: do we assume that joint probabilities are just the products of probabilities?

graemeblair:
Great. Yes, that's the variance we want I think.

I'm not sure, remind me where that's relevant? I haven't worked through all this since last fall.

acoppock:
The joint inclusion probabilities are the product of the marginal inclusion probabilities in simple, complete, and blocked designs (I *think* ????) but certainly not in clustered designs, since the joint inclusion probs for units in the same cluster is equal to the marginals in that case.

It’s probably worth writing to peter to get the code that generates table 2 from the data in table 1.

lukesonnet:
Right, the only joint inclusion probs you need to calculate for clustered designs are cross-cluster, as you treat can clusters as one for the variance estimation (at least as far as I can tell from eqn 35).

I can't see how the join inclusion probability = product of marginals is true in blocked designs, well at least in pair-matched designs.

I was also going to suggest writing to Peter. Who should?

[...]

PR #86, opened 2018-01-20
Work on how HT and DiM work with multi-valued treatments and expand scope of HT

lukesonnet:
This PR handles a few central issues in the two difference estimators.

-------
What this PR does
-------

**Conditions:** This PR updates how we work with multi-valued treatments in DiM and HT. It used to silently pick the two lowest treatment values and compare the second-smallest to the smallest. Now if there are multiple treatments, it forces you to specify the comparison between "condition2" and "condition1". Furthermore, if the treatment isn't 0/1, we append the value of condition2 to the treatment name in the coefficient_name output. For example if you pass:

```
dat <- data.frame(y = rnorm(4), z = c("T", "C"))
difference_in_means(y ~ z, condition2 = "T", data = dat)
```
It will return the coefficient name as "zT" and do the correct comparison.

**Horvitz-Thompson:** Thanks to @macartan and @acoppock, I learned that HT estimation was possible with no variation in treatment status. I implemented this, as well as the ability to do the other estimations posted the other day in Issue #18. They now work with a major exception which I will post below in a new comment. 

Furthermore, we now have assume `mean(treated)` and simple random assignment in HT when no information about the design is give. This is noted to the user in a warning, which I believe is appropriate.

I added tests for all of the above.

------
Outstanding questions
------

@acoppock, currently I know nothing about degrees of freedom in HT, and neither do any of the publications we've referenced. This leads to problems, were I can give you an estimate and a standard error if you give me only one observation, but I can't imagine any degrees of freedom would be correct. Not sure how we should handle this for now.

In general, there's something tricky about the conditions. First problem, if you pass a factor, you can't specify the conditions using the integers, you have to use the actual levels. I think this is okay. Second problem, if the user passes a vector of all 0s as the treatment for HT, and there are no treated units, and doesn’t pass condition1 or condition2, our default behavior is to treat the 0s as "treated". I think this is the correct behavior that allows us to be consistent by treating any treatments that are only one value as the "treated" unless they specify condition1 = 0. Any opinions?


lukesonnet:
@nfultz, the relevant code changes here are before commit 1ffa9ae, although looking at it altogether might be useful!

lukesonnet:
This PR has gotten *clobbered* by my other PRs. Basically the main files that need to be reviewed are:

* estimatr_horvitz_thompson.R
* `parse_conditions()` in helper_parse_arguments.R (and where it is called in estimatr_horvitz_thompson.R and estimatr_difference_in_means.R

PR #88, opened 2018-01-21
Builds support for learning from blocked designs for HT

lukesonnet:
This PR does three things.

**Allow users to get joint and marginal condition probabilities from blocked randomizr declarations:**  The function `declaration_to_condition_pr_mat()` now allows for blocked, blocked-clustered, and custom cases. I also create a series of tests that compare the condition probability matrix from this function with one from an arbitrary set of permutations (in the function `permutations_to_condition_pr_mat()`) to ensure that the resulting matrix is correct (in most tests I get the set of permutations from `randomizr::obtain_permutation_matrix()`). This has been done for all of the designs we currently support.

**Error learning from permutations:** It also fixes an error in `permutations_to_condition_pr_mat()` that put the treated condition in the upper part of the `condition_pr_mat()` that was not caught until now because the original test had a treatment probability of 0.5.

Lastly I cleaned up some other tests.


lukesonnet:
I merged in the newest PR (#89), because I realized it was getting too messy. @nfultz, when you look at the code, I recommend ignoring "Reorganizing R scripts" and later commits to avoid the glut of changed files (sorry about that). In those later commits, the only thing I really do is move creating the p-values and confidence intervals to a new script and add an error for any negative or zero weights.

tests/testthat/test-difference-in-means.R
```diff
@@ -325,6 +325,8 @@ test_that("DIM unbiased", {
   declaration <- randomizr::declare_ra(N = nrow(dat))
   treatment_perms <- randomizr::obtain_permutation_matrix(declaration)
 
+  treatment_perms
```
nfultz: dangler

R/helper_clean_model_data.R
```diff
@@ -97,6 +97,9 @@ clean_model_data <- function(formula,
 
   if(!missing(weights)){
     ret[["weights"]] <- model.extract(mf, "weights")
+    if (any(ret[["weights"]] <= 0)) {
+      stop("weights must all be positive and non-zero.")
```
nfultz: Zero weights should work?

tests/testthat/test-horvitz-thompson.R
```diff
@@ -184,6 +205,48 @@ test_that("Horvitz-Thompson properly checks arguments", {
     "N|declaration"
   )
 
+  # Reserved variable names
+  dat$`.clusters_ddinternal` <- 1
```
nfultz: For internal type things, also consider making them attributes instead of list elements. This will make pretty printing, iteration, etc easier

R/helper_clean_model_data.R
lukesonnet: I agree, was in response to @jaspercooper's comments. I think I'll change to < 0.

lukesonnet:
Will merge once linux and windows pass on the incorporated changes.

PR #89, opened 2018-01-21
Restructures a bunch of files; adds tests and errors

lukesonnet:
This PR restructures/renames a lot of the files and adds tests in some weak spots. It also throws errors for some poorly made args and creates a function for some common usage across estimators (getting p-values and confidence intervals).


tests/testthat/test-horvitz-thompson.R
```diff
@@ -184,6 +205,48 @@ test_that("Horvitz-Thompson properly checks arguments", {
     "N|declaration"
   )
 
+  # Reserved variable names
+  dat$`.clusters_ddinternal` <- 1
```
nfultz: dollar backtick is ugly, let's use `dat[[".clusters_ddinternal"]]`

PR #97, opened 2018-01-24
Compatibility with margins

lukesonnet:
This PR seeks to make sure we work with leeper/margins.

So far, so good! The main function from that package `margins::margins` which estimates average marginal effects and uses the delta method or bootstrapping works with our package. His summary and margins methods work because as far as I can tell Neal and I have gotten it everything it needs!

It also works with `lm_lin`.

There are two main catches:

1. The old version (on CRAN) of his package requires you to have `"coefficients"` in your outcome instead of relying on `coef(.)`, which his github version does. For now I just duplicate `est` into `coefficients` but we should probably just replace `est` with `coefficients` across all estimators as its more standard and this might not be the last time this comes up. I changed `coef.lm_robust()` to also return `coefficients` because the old version of the package just clobbers the coefficients to get new predictions for the delta method.

2. It doesn't deal super gracefully with us having a "clusters" column. It creates a messy column in `margin_effects(lmout)` and throws a lot of warnings when you call `margins(lmout)`, but it still works!

-----
If you review this, ignore the first 3 commits (sorry), they are old doc things I noticed. I'm too close to release to not fix a problem as soon as I see it.

Quick decisions:
- replace est with coefficients package wide? I vote yes.
- New vignette? not sure where it belongs but I'd love to write this up. It could just be a line in getting-started. But I'd like to start a vignette that eventually would also show integration w texreg.


nfultz:
Lets add another vignette in the next version.

lukesonnet:
Ok as soon as this works on windows + linux and Neal thinks its good, I'm merging it.

R/S3_print.R
```diff
@@ -8,7 +8,57 @@ print.lm_robust <-
...
+#' @param ... arguments passed to \code{\link{tidy.lm_robust}}, unused
+#'
+#' @export
+print.summary.lm_robust <-
+  function(
+    x,
+    digits = max(3L, getOption("digits") - 3L),
+    ...) {
+
+    cat("\nCall:\n", paste(deparse(x$call), sep = "\n", collapse = "\n"),
```
nfultz: please set nlines here and other places we might use deparse.

PR #103, opened 2018-01-25
Bug fix in cond_pr_mat; trimming docs

lukesonnet:
Fixes a pretty big bug in helper_condition_pr_mat.R. going back over each block again to create the cross-block probabilities was happening too early (some blocks hadn't had the probabilities added yet!)

Also fixed win-builder error and trimmed some docs.


R/helper_clean_model_data.R
```diff
@@ -107,7 +107,7 @@ clean_model_data <- function(formula,
 
     if (any(ret[["condition_pr"]] <= 0 | ret[["condition_pr"]] > 1)) {
       stop(
-        "`condition_pr` must be a vector of positive values no greater than 1"
+        "`condition_prs` must be a vector of positive values no greater than 1"
```
nfultz: does not match the code ??? formal argument does not have the s

tests/testthat/test-horvitz-thompson.R
```diff
@@ -203,7 +206,7 @@ test_that("Horvitz-Thompson works with clustered data", {
   dat$p_wrong[1] <- 1.5
   expect_error(
     horvitz_thompson(y ~ z, data = dat, clusters = cl, condition_prs = p_wrong),
-    "`condition_pr` must be a vector of positive values no greater than"
+    "`condition_prs` must be a vector of positive values no greater than"
```
nfultz: no greater than what?

R/helper_clean_model_data.R
lukesonnet: that's because clean_model_data doesn't use plural. It should. The `condition_prs` matches the argument users see.

tests/testthat/test-horvitz-thompson.R
lukesonnet: In a lot of my tests I don't paste full errors :( its 1

R/helper_clean_model_data.R
lukesonnet: I don't think I want to fix this right now? It's all internal (but problematic).

lukesonnet:
Merging so we can just run the tests on master as it passes on local.

PR #132, opened 2018-02-14
Make tweaks for CRAN

lukesonnet:
There were some "deep" CRAN errors we needed to resolve that you can see here: https://cran.r-project.org/web/checks/check_results_estimatr.html

1. Tests failed without Long Doubles (https://www.stats.ox.ac.uk/pub/bdr/noLD/estimatr.out). They were not identical but rather equal, so I just changed the test.
2. Uninitialized values (the degrees of freedom, residuals, and residual variance in `lm_robust_helper.cpp` were being used and returned to the R wrapper code and were causing errors in Valgrind. I now initialize these values and tested all of the test files one by one using valgrind and there are no more errors.
3. Solaris errored on a test that `expect_equal` the max diff of two `margins` was less than < 2e05. I changed the test to `expect_equals` with a tolerance of 0.01. I have no way to test this since rhub's Solaris platform apparently cannot compile RcppEigen files (and I'm not sure if I can fix this). See r-hub/rhub/issues/109.

I think if this passes, we might as well try to resubmit and see what they say?

Oh, I also got rid of my silly way to handle if the cholesky decomposition failed that relied on catching my own error. Logic should be much cleaner now.

### DIFF
--- a/src/lm_robust_helper.cpp
+++ b/src/lm_robust_helper.cpp
@@ -32,24 +32,23 @@ List lm_solver(Eigen::Map<Eigen::MatrixXd>& Xfull,
   Eigen::MatrixXd XtX_inv, R_inv, Vcov_hat;
   Eigen::VectorXd beta_out(Eigen::VectorXd::Constant(p, ::NA_REAL));
 
-  try {
-    if (try_cholesky) {
-      const Eigen::LLT<Eigen::MatrixXd> llt(Xfullt*Xfull);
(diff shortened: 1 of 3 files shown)


lukesonnet:
Sorry @nfultz , I'm not sure where a lot of the diffs in the cpp came from, most are useless.

nfultz:
LGTM, feel free to merge

PR #133, opened 2018-02-14
Prep for CRAN resubmission

lukesonnet:


nfultz:
also don't forget to tag the release when you merge / submit - I recommend uploading the .tar.gz that travis puts in drat just to keep everything consistent.

PR #140, opened 2018-02-20
Update DESCRIPTION

nfultz:
Bumping versions required on imports / suggests


lukesonnet:
Let's merge once these pass? Just want to make sure it can actually get these versions.

nfultz:
Might need additional_repositories set.

lukesonnet:
It;s already set

PR #141, opened 2018-02-21
Transfer NSE handling to rlang; implement IV estimator; refactor lm_robust_helper.cpp

lukesonnet:
This is a doozy of a PR. Each of the following bullet points is a major change:

- refactor cpp to have separate LS solutions and variance estiamtion
- refactor clean_model_data to use rlang
- add support for multiple outcome variables
- add instrumental variables


R/estimatr_difference_in_means.R
```diff
@@ -215,25 +215,22 @@ difference_in_means <-
            condition2 = NULL,
            ci = TRUE,
            alpha = .05) {
-    if (length(all.vars(formula[[3]])) > 1) {
+    if (length(all.vars(eval_tidy(formula)[[3]])) > 1) {
```
nfultz: suggest using f_rhs hear instead of [[3]] - little easier to read.

R/estimatr_horvitz_thompson.R
```diff
@@ -283,8 +283,8 @@ horvitz_thompson <-
       # Add clusters, blocks, and treatment probabilities to data so they can be cleaned with clean_model_data
       if (!is.null(declaration$clusters)) {
         if (is.null(data[[".clusters_ddinternal"]])) {
-          data[[".clusters_ddinternal"]] <- declaration$clusters
-          clusters <- ".clusters_ddinternal"
+          .clusters_ddinternal <- declaration$clusters
+          clusters <- quo(.clusters_ddinternal)
```
nfultz: Flip the test and deindent.

R/helper_clean_model_data.R
```diff
@@ -85,25 +101,25 @@ clean_model_data <- function(formula,
     ret[["original_treatment"]] <- mf[, colnames(mf) == all.vars(terms(mf)[[3]])[1]]
   }
 
-  if (!missing(weights)) {
+  if (!is.null(mfargs$weights)) {
```
nfultz: is.numeric

tests/testthat/test-iv-robust.R
```diff
@@ -0,0 +1,156 @@
+context("Estimator - iv_robust")
+
+skip_if_not_installed("AER")
```
nfultz: skip / require a package inside the specific test_that block

nfultz: idk

src/lm_robust_helper.cpp
```diff
@@ -14,44 +14,46 @@ Eigen::MatrixXd AtA(const Eigen::MatrixXd& A) {
...
-               const std::vector<bool> & which_covs,
+Eigen::MatrixXd Kr(const Eigen::MatrixXd& A, const Eigen::MatrixXd& B) {
+  Eigen::MatrixXd AB(A.rows() * B.rows(), A.cols() * B.cols());
+
+  for (int i = 0; i < A.rows(); i++) {
+    for (int j = 0; j < A.cols(); j++) {
+      AB.block(i*B.rows(), j*B.cols(), B.rows(), B.cols()) = A(i, j) * B;
+    }
+  }
+  return AB;
```
nfultz: somewhat surprised you had to roll your own kronecker product.

src/lm_robust_helper.cpp
```diff
@@ -106,349 +108,395 @@ List lm_solver(Eigen::Map<Eigen::MatrixXd>& Xfull,
...
+          (Eigen::MatrixXd::Identity(len, len) - H) - H.transpose() +
+            Xoriginal.block(start_pos, 0, len, r) *
+            MUWTWUM *
+            Xoriginal.block(start_pos, 0, len, r).transpose()
+      );
+
+      Eigen::VectorXd eigvals = At_WX.eigenvalues();
+      for (int m = 0; m < eigvals.size(); ++m) {
+        if (eigvals(m) > std::pow(10.0, -12.0)) {
+          eigvals(m) = 1.0 / std::sqrt(eigvals(m));
```
nfultz: :(

nfultz: operatorInverseSqrt ?

lukesonnet: No because it fails when the eigenvalues are 0 or negative :/

src/lm_robust_helper.cpp
lukesonnet: Unsupported 🙌 https://eigen.tuxfamily.org/dox/unsupported/group__KroneckerProduct__Module.html

R/estimatr_horvitz_thompson.R
lukesonnet: I'm not sure I follow. This makes sure we only warn in appropriate cases?

lukesonnet:
Very sorry about the size of this one :(. Each new thing necessitated another new thing that I didn't foresee and it blew up.

R/helper_clean_model_data.R
lukesonnet: Actually, no on this one because the arg isn't the weights itself, its just the name of the weights object that is evaluated above.

lukesonnet: I rescind my thumbs up!

R/estimatr_horvitz_thompson.R
nfultz:
```r
if(test){
  stop()
}
# This part not in an else block (bc stop above exits), not indented.
```

R/helper_clean_model_data.R
nfultz: If weights is not in the model.frame, it returns NULL, and    
```
  ret[["weights"]] <- model.extract(mf, "weights")
```
still works unconditionally.

R/estimatr_horvitz_thompson.R
lukesonnet: Then we have to check a condition twice.

lukesonnet:
```
      if (!is.null(declaration$clusters) && !is.null(data[[".clusters_ddinternal"]])) {
        stop(
          "Can't have a variable called '.clusters_ddinternal' in your `data`",
          "as we use that name for clusters from `declaration` objects."
        )
      }
      if (!is.null(declaration$clusters)) {
        .clusters_ddinternal <- declaration$clusters
        clusters <- quo(.clusters_ddinternal)
      }
```

R/helper_clean_model_data.R
nfultz:
```
> ret <- list()
> mf <- model.frame(extra~group, data=sleep)
> ret[["weights"]] <- model.extract(mf, "weights")
> ret
list()
> any(ret[["weights"]] < 0)
[1] FALSE
```

R/estimatr_horvitz_thompson.R
nfultz:
```
      if (!is.null(declaration$clusters) {
        if(!is.null(data[[".clusters_ddinternal"]])) {
         stop(
            "Can't have a variable called '.clusters_ddinternal' in your `data`",
            "as we use that name for clusters from `declaration` objects."
          )
        }
        .clusters_ddinternal <- declaration$clusters
        clusters <- quo(.clusters_ddinternal)
      }

Comment (lukesonnet) [reply R/estimatr_horvitz_thompson.R]: Ugh I was stuck on reworking the whole conditional sorry

File: src/lm_robust_helper.cpp (hunk 6 above)
Comment (nfultz) [reply src/lm_robust_helper.cpp]: kk if eigen doesn't handle that.

For the next version,  I think we should pull this block out to it's own helper function eg `psuedoinverse()` / `generalizeinverse()`.

* early exit from the for loop, eigenvalues are sorted descending.

* instead of 1 / sqrt use pow(-.5) which should have better precision for values close to 0

* instead of leaving zeros in and multiplying the full matrix by zeros, you know how many eigenvalues are non-zero (from early exit above), so slice the V matrix to that many columns before the multiplication. If the matrix is large and low rank this is could be a decent constant to shave.

Final calling code could look like:

```
 Eigen::MatrixXd At_WX_inv = pseudoinverse(
          (Eigen::MatrixXd::Identity(len, len) - H) - H.transpose() +
            Xoriginal.block(start_pos, 0, len, r) *
            MUWTWUM *
            Xoriginal.block(start_pos, 0, len, r).transpose()
      );
```

Also if H is symmetric you can omit the transpose or scalar multiply instead.

lukesonnet: Thanks, I think that's right.

Re:H, it is not symmetric if eights are applied, which is why there is a warning on the packages front page from the last time I removed this transpose.

lukesonnet:
Enough have passed, I'm merging so you can go ahead!

PR #143, opened 2018-02-21
Nfultz/rlang

nfultz:


R/helper_clean_model_data.R
```diff
@@ -1,24 +1,47 @@
...
+  m_formula <- eval_tidy(mfargs[["formula"]])
+  m_formula_env <- environment(m_formula)
+
+  # For each ... that would go to model.fram .default, early eval, save to formula env, and point to it
+  # subset is also non-standard eval
+  to_process <- setdiff( names(mfargs), setdiff( names(formals(stats::model.frame.default)),"subset") )
+
+  for (da in to_process) {
+    name <- sprintf(".__%s%%%d__", da, sample.int(.Machine$integer.max, 1))
+    m_formula_env[[name]] <- eval_tidy(mfargs[[da]], data = data)
```
lukesonnet: I think my only question is why you need to evaluate the arguments within the formula's environment. Is that because this is where `model.frame` will look if not in data?

nfultz: exactly, see also https://github.com/wch/r-source/blob/master/src/library/stats/R/models.R#L390-L428

lukesonnet: Great, and the name is just to avoid collisions. Can't we just remove the sample with some large, arbitrary integer?

nfultz: Let me take another pass to try to clean that up - my fear is that people reuse a formula object across many models, so sneaking data through the formula environment could lead to corruption if model.frame is doing any other shenanigans I'm unaware of.

PR #163, opened 2018-03-08
Try apt install nlopt

lukesonnet:


nfultz:
LGTM but might make the mac build fail? Worth a try.

lukesonnet:
Failed on ubuntu anyways.... I can download the tarball on my machine, but travis obviously is failing to get it. Will keep exploring

nfultz:
Also worth noting that the host for travis is super old (ubutun 12 or 14
iirc)

Issue #148, opened 2018-03-05
lin estimator should weight before scaling


lukesonnet:

lukesonnet:
I believe I have this implemented appropriately now. The following test passes:

```r
  lmwo <- lm_lin(mpg ~ am, ~ hp, weights = wt, data = mtcars)
  hp_wmean <- weighted.mean(mtcars$hp, mtcars$wt)
  wareg <- lm(mpg ~ hp, data = mtcars, subset = am == 1, weights = wt)
  wbreg <- lm(mpg ~ hp, data = mtcars, subset = am == 0, weights = wt)
  wate <-
    with(mtcars[mtcars$am == 1, ],
         weighted.mean(mpg, wt) + (hp_wmean - weighted.mean(hp, wt)) * wareg$coefficients[2]) -
    with(mtcars[mtcars$am == 0, ],
         weighted.mean(mpg, wt) + (hp_wmean - weighted.mean(hp, wt)) * wbreg$coefficients[2])

  expect_equivalent(
    wate,
    lmwo$coefficients["am"]
  )
```

However, the following is no longer true:
```r
  dat <- data.frame(Y = rnorm(4))
  dat$bl <- rep(1:25, each = 16)
  dat$z_blocked <- rep(c(0, 1), each = 2)
  dat$w <- runif(nrow(dat))
  expect_equivalent(
    tidy(lm_lin(Y ~ z_blocked, ~ factor(bl), weights = w, data = dat))[2, ],
    tidy(difference_in_means(Y ~ z_blocked, data = dat, blocks = bl, weights = w))
  )
```

I think this is because I'm not properly combining weighted block means. Will investigate right now.

lukesonnet:
Closed by #169

PR #169, opened 2018-03-13
Fixes weighting for lm_lin and blocked difference_in_means

lukesonnet:
Previously, `lm_lin` was not demeaning covariates by the weighted mean, and was just using the simple mean (Issue #148). This has been solved.

After solving it, a test failed checking `difference_in_means` against `lm_lin` in the blocked, weighted case. This test previously passed because both were wrong in the same say!

The above problem is solved now. In `difference_in_means()` we compute the post-stratification estimator as:

```r
diff <- with(block_estimates, sum(coefficients * N / N_overall))
```

The problem was that we used `N` and `N_overall` rather than the sum of the weights in a block and the sum of all of the weights if you had a blocked, weighted specification. Now `N` and `N_overall` appropriately capture the weights.

### DIFF
--- a/R/estimatr_difference_in_means.R
+++ b/R/estimatr_difference_in_means.R
@@ -238,7 +238,11 @@ difference_in_means <-
       stringsAsFactors = FALSE
     )
     data$cluster <- model_data$cluster
-    data$weights <- model_data$weights
+    # rescale weights for convenience
+    if (is.numeric(model_data$weights)) {
+      weight_mean <- mean(model_data$weights)
+      data$weights <- model_data$weights / weight_mean
+    }
     data$block <- model_data$block
 
     if (!is.null(data$weights) && length(unique(data$weights)) == 1
@@ -355,7 +359,7 @@ difference_in_means <-
         ## matches lm_lin, two estimates per block
         if (is.null(data$cluster)) {
           design <- "Blocked"
-          df <- N_overall - 2 * n_blocks
+          df <- nrow(data) - 2 * n_blocks
         } else {
           design <- "Block-clustered"
           # Also matches lm_lin for even sized clusters, should be conservative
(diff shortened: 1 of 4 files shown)


R/estimatr_lm_lin.R
```diff
@@ -253,14 +253,25 @@ lm_lin <- function(formula,
...
+      ,
+      setdiff(colnames(design_matrix), c(design_mat_treatment, "(Intercept)")),
+      drop = FALSE
+    ]
+
+  # Choose what to center on!
+  if (is.numeric(weights)) {
+    center <- apply(demeaned_covars, 2, weighted.mean, weights)
+  } else {
+    center <- TRUE
```
nfultz:
```
center <- colMeans(...)
```

or simliar - generally don't set a variable to different types in different branches.

R/estimatr_lm_lin.R
```diff
@@ -253,14 +253,25 @@ lm_lin <- function(formula,
...
   demeaned_covars <-
     scale(
-      design_matrix[
-        ,
-        setdiff(colnames(design_matrix), c(design_mat_treatment, "(Intercept)")),
-        drop = FALSE
-      ],
-      center = TRUE,
+      demeaned_covars,
+      center = center,
```
nfultz: Given above, change scale to sweep.

tests/testthat/test-lm-lin.R
```diff
@@ -311,3 +311,54 @@ test_that("Test LM Lin", {
...
+         mean(mpg) + (m_hp - mean(hp)) * areg$coefficients[2]) -
+    with(mtcars[mtcars$am == 0, ],
+         mean(mpg) + (m_hp - mean(hp)) * breg$coefficients[2])
+
+  expect_equivalent(
+    ate,
+    lmo$coefficients["am"]
+  )
+
+  # Weighted matches (one covar)
```
nfultz: split these into different test_that blocks with appropriate lables.

R/estimatr_difference_in_means.R
```diff
@@ -238,7 +238,11 @@ difference_in_means <-
       stringsAsFactors = FALSE
     )
     data$cluster <- model_data$cluster
-    data$weights <- model_data$weights
+    # rescale weights for convenience
+    if (is.numeric(model_data$weights)) {
+      weight_mean <- mean(model_data$weights)
+      data$weights <- model_data$weights / weight_mean
```
nfultz: If weight_mean is never used again, ok to put all onto one line.

PR #172, opened 2018-03-19
Errors/warnings and docs for IV

lukesonnet:


R/helper_lm_robust_fit.R
```diff
@@ -157,8 +157,7 @@ lm_robust_fit <- function(y,
     return_list <- list(
       coefficients = setNames(as.vector(fit$beta_hat), variable_names),
       se = NA,
-      df = NA,
-      stringsAsFactors = FALSE
+      df = NA
```
nfultz: Do you also need to cut stringsasfactors if(multivariate) is true?

tests/testthat/test-iv-robust.R
```diff
@@ -106,6 +117,123 @@ test_that("iv_robust matches AER + ivpack", {
...
+
+  expect_equal(
+    ivdefclr$coefficients,
+    ivdefcl$coefficients
+  )
+
+  expect_equivalent(
+    as.matrix(tidy(ivdefclr)[1:2, c("coefficients", "se")]),
+    ivdefclse[, c(1, 2)]
+  )
```
nfultz: Suggest making a helper function that does the repeated expect_equals for you - eg 

```r
    expect_equal_models <- function(model1, model2) {
      expect_equal(model1$coefficients, model2$coefficients)
      expect_equal(tidy(model1$coefficients), tidy(model2$coefficients))


```


etc.

R/estimatr_iv_robust.R
```diff
@@ -71,6 +126,18 @@ iv_robust <- function(formula,
   data <- enquo(data)
   model_data <- clean_model_data(data = data, datargs)
 
+  if (is.null(model_data$instrument_matrix)) {
+    stop(
+      "Must specify a `formula` with both regressors and instruments. For ",
+      "example, `formula = y ~ x1 + x2 | x1 + z2` where x1 and x2 are the ",
+      "regressors and z1 and z2 are the instruments.\n\nSee ?iv_robust."
+    )
+  }
```
nfultz: Is there someway to write this error message so that it checks the formula (an input) rather than the output of clean model data.

R/helper_lm_robust_fit.R
lukesonnet: This used to be a df, now it's a list. So there's no case where I need that as an item in the list, right?

R/estimatr_iv_robust.R
lukesonnet: Sure, but this is what implicitly happens in clean_model_data, and I'm a little worried about making sure I'm evaluating the formula arg properly before testing it

nfultz: kk, suggest moving this error check into clean_model_data then. the output of clean_model_data should (if it succeeds) not need typically need further testing / checking.

R/estimatr_lm_lin.R
```diff
@@ -181,7 +184,7 @@ lm_lin <- function(formula,
   # Check covariates is right hand sided fn
   if (attr(cov_terms, "response") != 0) {
     stop(
-      "`covariates` must be right-sided formula only, such as '~ x1 + x2 + x3'"
+      "The `covariates` argument must be right-sided formula only, such as '~ x1 + x2 + x3'"
```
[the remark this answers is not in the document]
lukesonnet: Thanks, gonna keep the recommendation as well

tests/testthat/test-iv-robust.R
lukesonnet: It's a good idea given how much of this there is, but the main problem is depending on my comparison I use different columns I expect to be the same, and they are named different things across packages.

R/estimatr_iv_robust.R
lukesonnet: I don't know why I would be averse to internal arguments.... but I was. This makes sense.

PR #173, opened 2018-03-19
Changes tidy output

lukesonnet:
Really weird that the list output doesn't match the tidy output now, but I think matching the tidy output is probably right so that everything is internally consistent.. There are 3 problems with also changing the list output of the estimator functions:

1. `lm` returns an object with `res$coefficients` and if we change our list so that the coefficients are in `robres$estimate`, but `summary(robres)$coefficients` still matches `summary(res)$coefficients`, then we are sometimes matching, sometimes not.
2. We have a (small) set of users who might be thrown off by this (I don't think this is a real big concern as now is the time to make this change if ever).
3. We have to rework DD diagnosands and designs to also be consistent, because otherwise it's still internally inconsistent

I think @nfultz is in favor of aligning everything, doing the work across DD and estimatr, and just not worrying about matching `lm` output.

I agree with him.

I need input from the bosses so that we can fix this today and go for CRAN with something we are committed to for the long run.

### DIFF
--- a/R/S3_summary.R
+++ b/R/S3_summary.R
@@ -103,7 +103,7 @@ summary.horvitz_thompson <-
   }
 
 summarize_tidy <- function(object, test = "t", ...) {
-  remove_cols <- c("coefficient_name", "outcome")
+  remove_cols <- c("term", "outcome")
 
   # This is ugly SO THAT summary(fit)$coefficients returns something like lm does.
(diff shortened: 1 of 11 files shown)


nfultz:
This PR only changes tidy() and runs cleanly against the DeclareDesign master - go ahead and merge this in and start a new branch for changing output model structure.

PR #179, opened 2018-03-19
rmcall failures

nfultz:
Fixes tests failing like

test-lm-cluster.R:286: error: lm works with quoted or unquoted vars and withor without factor clusters
could not find function "rmcall"
1: expect_equivalent(rmcall(lm_robust(Y ~ Z, data = dat, clusters = J_fac)), rmcall(lm_robust(Y ~ Z, data = dat, 
       clusters = J))) at /home/nfultz/projects/DD/estimatr/tests/testthat/test-lm-cluster.R:286
2: quasi_label(enquo(object), label)
3: eval_bare(get_expr(quo), get_env(quo))

### DIFF
--- a/tests/testthat/helper-rm-call.R
+++ b/tests/testthat/helper-rm-call.R
@@ -1,7 +1,0 @@
-# This file removes calls from function returns to make testing easier
-rmcall <- function(obj) {
-  if (!is.null(obj[["call"]])) {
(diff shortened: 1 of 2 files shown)


lukesonnet:
rmcall is used in multiple files, not just lm_robust, hence the failure and why it is in a helper. Should I add this to every tester file it is used in?

PR #186, opened 2018-03-20
estimate -> coefficients in model object

lukesonnet:


tests/testthat/test-lm-lin.R
```diff
@@ -84,7 +84,7 @@ test_that("Test LM Lin", {
   )
 
   expect_equivalent(
-    lm_lin_out$estimate,
+    lm_lin_out$coefficients,
```
nfultz: In your tests, use `coef(lm_lin_out)` instead of `lim_lin_out$coefficients`

lukesonnet: oh boy

PR #190, opened 2018-03-27
Workaround and error for IV where fstat cannot be computed

lukesonnet:
Happy to hear better solutions to this problem.


lukesonnet:
Should we not try to get this on to CRAN in this submission?

nfultz:
No, I think this needs to wait for the next release instead of restarting the last release cycle over again.

lukesonnet:
Ok.

PR #193, opened 2018-03-28
Fixes IV doc; changes warning

lukesonnet:
Also updates website


R/helper_condition_pr_matrix.R
```diff
@@ -131,15 +138,15 @@ declaration_to_condition_pr_mat <- function(ra_declaration,
   declaration_call <- as.list(ra_declaration$original_call)
   simple <- eval(declaration_call$simple)
 
-  n <- nrow(ra_declaration$probabilities_matrix)
+  n <- nrow(prob_matrix)
 
-  if (ra_declaration$ra_type == "simple") {
+  if ("ra_simple" %in% class(ra_declaration)) {
```
nfultz: use inherits() here instead.

nfultz: also consider factoring out to generic helper function,.

PR #205, opened 2018-05-06
PR for active stargazer integration

### LINKED ISSUE #116: moving commarobust() to estimatr
We're going to move the function commarobust() from the commarobust package. 

It should take an lm object and return tidy df with robust standard errors that looks exactly the same as lm_robust. 

We're going to deprecate the functionality of the commarobust package and push people to this.

### LINKED ISSUE #216: Include var names in standard errors from lm_robust
```
dat <- fabricate(
  N = 40,
  y = rpois(N, lambda = 4),
  x = rnorm(N),
  z = rbinom(N, 1, prob = 0.4)
)
lmro <- lm_robust(y ~ x + z, data = dat)


# Coefficients have names
lmro$coefficients

# but std.errors don't
lmro$std.error
```

lukesonnet:
This PR:

- Adds `commarobust` and `starprep` (closes #116)
- Refactors quite a lot of `lm_robust_fit` in the process of doing the above
- Harmonizes output across objects in `lm_robust` output (closes #216)
- Makes some minor web changes


R/starprep.R
```diff
@@ -0,0 +1,26 @@
...
+#'
+#'
+starprep <- function(..., stat = "se"){
+  fitlist = list(...)
+  if(stat == "se") {
+    out <- lapply(fitlist, function(x) x$std.error)
+  } else if (stat == "p") {
+    out <- lapply(fitlist, function(x){
+      ps <- x$p.value
+      ps["(Intercept)"] <- 1
```
lukesonnet: @acoppock, why is the p-value on the intercept always one?

tests/testthat/test-iv-robust.R
```diff
@@ -163,7 +163,7 @@ test_that("iv_robust matches AER + ivpack", {
   dat$x1_c <- dat$x
   ivdefr <- iv_robust(y ~ x + x1_c| z + z2, data = dat, se_type = "HC0")
   ivdef <- ivreg(y ~ x + x1_c| z + z2, data = dat)
-  ivdefse <- robust.se(ivdef)
+  capture_output(ivdefse <- robust.se(ivdef))
```
nfultz: why all the capture outputs?

nfultz: shouldn't need them.

R/helper_lm_robust_fit.R
```diff
@@ -272,54 +226,54 @@ lm_robust_fit <- function(y,
   if (return_fit) {
     if (iv_second_stage && !fes) {
       if (weighted) {
-        return_list[["fitted.values"]] <- as.matrix(X_first_stage_unweighted %*% fit$beta_hat)
+        return_list[["fitted.values"]] <- as.matrix(data[["X_first_stage_unweighted"]] %*% fit$beta_hat)
       } else {
         return_list[["fitted.values"]] <- as.matrix(fitted.values)
       }
-    } else if ((se_type == "CR2" && weighted)) {
+    } else if (se_type == "CR2" && weighted && (iv_first_stage || !return_unweighted_fit)) {
```
nfultz: This logic is confusing, roughly six different flags are interacting. Add a todo to refactor after this next version ships.

tests/testthat/test-iv-robust.R
lukesonnet: ivpack prints whether the SEs are clustered or just robust w/ every call.

R/helper_lm_robust_fit.R
lukesonnet: Yes, this is absolutely brutal.

tests/testthat/test-iv-robust.R
nfultz: https://github.com/cran/ivpack/blob/master/R/robust.se.R#L4

🤦‍♂️

lukesonnet: Yep.

R/helper_starprep.R
```diff
@@ -99,6 +99,7 @@ commarobust <- function(model, se_type = NULL, clusters = NULL, ci = TRUE, alpha
     }
   }
 
+  print(data[["cluster"]])
```
nfultz: ?

lukesonnet: uhhhhh

lukesonnet:
I did a little bit of a refactor. I think it's more or less a reorganization, but it's certainly more readable now.

lukesonnet:
@graemeblair or @nfultz, please take a look if you can.

R/helper_lm_robust_fit.R
```diff
@@ -241,9 +241,9 @@ lm_robust_fit <- function(y,
 
       }
     } else if (weighted) {
-      return_list[["fitted.values"]] <- fit_vals[["fitted.values.unweighted"]]
+      return_list[["fitted.values"]] <- as.matrix(fit_vals[["fitted.values.unweighted"]])
```
nfultz: This reads as "if it's weighted, return the unweighted fits" ?

lukesonnet: Yes, we always return unweighted fits.

nfultz:
Rereading lm_robust_fit, it seems like there are three iv "modes" now - 
* no iv
* iv 1st stage
* iv 2nd stage

where iv 2nd stage is assumed if X_first_stage is passed in. 


Instead of two essentially binary arguments, (eg could be both 1st stage and 2nd stage) suggest refactoring function argument to

```
lm_robust_fit <- function(..., stage=NULL)
```

and in iv_robust, pass it an object such as `stage=list(1)`, `stage=list(2, X_first_stage)`.

This would also make understanding this test easier:

```
if (fes && !iv_first_stage)
```

It's for no-iv and iv 2nd stage I guess?

lukesonnet:
That's right, and a good point. I can implement this for sure.

lukesonnet:
We still would have to create an IV flag, and it would double the checks, then, as we would have to do && iv && stage[[1]] == 2, so I'm actually not sure it's much better.

nfultz:
Why would we need an extra iv flag in the function arguments? if stage ==
FALSE then don't do the iv .


```
> stage <- FALSE
>
> stage[1] == 2
[1] FALSE
>
> stage <- list(1)
>
> stage[1] == 2
[1] FALSE
>
>
> stage <- list(2, sleep)
>
> stage[1] == 2
[1] TRUE
```

lukesonnet:
Expecting an argument to take two different data structures?

nfultz:
You could make the default `list(0)`, it would still do the same thing.

lukesonnet:
@nfultz, @graemeblair, we're ready on this I think. Just need a co-sign.

PR #208, opened 2018-05-08
Fix some broken links and other web issues

lukesonnet:


lukesonnet:
I think that's only in the man pages, can't find it anywhere else. Fixed.

lukesonnet:
So I did that on the github. I didn't do that on the README because I think I was told a while ago that this would confuse users on the home page. I can hide it though using echo. (In Rmd you need to set seed within the same chunk)

lukesonnet:
I believe this is solved now.

PR #209, opened 2018-05-11
Fixed effects; fixes to f-stats and r2s

lukesonnet:
The main purpose of this PR is to implement fixed effects estimators that use demeaning to speed up estimation when there are many groups. Here is a summary of the changes:

1. Bugfix for `r.squared` and `adj.r.squared` when CR2 standard errors were estimated
2. Bugfix for all `fstatistic`s with robust standard errors
3. Added `fixed_effects` argument as a RHS formula for `lm_robust` and `iv_robust`

Some remaining decisions that come out of this PR:

1. The way we count the rank of the FEs for the DoFs and finite sample corrections is rudimentary for `"classical", "HC0", "HC1", "CR0", "stata"` standard errors. If normally some group effects would be dropped out due to intersecting groups or due to collinearity in the fixed effects, the fixed effects estimator does not do this. That means if there are collinearities in the fixed effects the standard errors from passing the FEs as dummies and passing them to `fixed_effects` will differ for the above standard error estimators.
2. The fixed effects estimator increases the probability that there are perfect fits (such as a group with only one unit). In these cases, we often can get some instability in the variance covariance matrix. [This line shows one way that I try to work around this problem, when there are some issues with numerical zeros](https://github.com/DeclareDesign/estimatr/pull/209/files#diff-fa3978062e1972b525e7008b94a2f4eaR306). This fix shouldn't break anything, but it is a bit hacky. Any thoughts?
3. The update to the calculation of `fstatistic`s means that in many cases, we can't compute the fstat as we have to invert a large chunk of the variance-covariance matrix. Currently this issues a warning, but I suppose we should instead just silently pass an NA to the `fstatistic` output. Do we agree?


acoppock:
On your decisions:

1 - the documentation for fixed effects should point people to things that will let them know if they are asking for bad fits.  You'd mentioned some function? But otherwise I think that's the right approach.
2 - I can't interpret that line... but I trust you?
3 - NA makes sense. But can't we get an f-stat for the "projected" data easily?

We need to do a pass through the S3 methods for FE models, e.g.:

```
rm(list = ls())
library(randomizr)

Y <- rnorm(4000)
blocks <- factor(rep(1:2000, each = 2))
Z <- block_ra(blocks = blocks)
dat <- data.frame(Y, Z, blocks)

fit_1 <- lm_robust(Y ~ Z + blocks, data = dat)
fit_2 <- lm_robust(Y ~ Z, fixed_effects = ~ blocks, data = dat)

# works great
predict(fit_1, newdata = data.frame(Y, Z, blocks))
# does not work!
predict(fit_2, newdata = data.frame(Y, Z, blocks))
```

R/helper_clean_model_data.R
```diff
@@ -31,7 +31,22 @@ clean_model_data <- function(data, datargs, estimator = "") {
 
   for (da in to_process) {
     name <- sprintf(".__%s%%%d__", da, sample.int(.Machine$integer.max, 1))
-    m_formula_env[[name]] <- eval_tidy(mfargs[[da]], data = data)
+    if (da == "fixed_effects") {
```
nfultz: Move this test / conversion below the loop, near the formula conversion. Also consider vapply instead of sapply.

R/helper_clean_model_data.R
```diff
@@ -129,3 +153,33 @@ clean_model_data <- function(data, datargs, estimator = "") {
...
+  )
+
+  # save names
+  dimnames(demeaned[["newY"]]) <- dimnames(model_data[["outcome"]])
+  new_names <- dimnames(model_data[["design_matrix"]])
+  new_names[[2]] <- new_names[[2]][new_names[[2]] != "(Intercept)"]
+  dimnames(demeaned[["newX"]]) <- new_names
+
+  model_data[["outcome"]] <- demeaned[["newY"]]
+  model_data[["design_matrix"]] <- demeaned[["newX"]]
```
nfultz: Ideally, match up the names on demeaned with model_data - eg demeaned$outcome instead of demeaned$newY etc.

lukesonnet:
Thanks, I'll be able to resolve most of this tomorrow, I hope. Yes we can easily get the projected F-statistic.

One more thing that I left out: it is likely that computing this F-statistic is pretty slow. I'll do some speed tests and see if it's bad enough that we need to come up with a solution.

tests/testthat/test-s3-methods.R
```diff
@@ -24,28 +24,55 @@ test_that("tidy, summary, and print work", {
...
 
   capture_output(
     summary(lmo)
   )
 
+  capture_output(
+    summary(lmfo)
+  )
+
+  expect_true(
```
nfultz: `testthat::expect_output`

lukesonnet: cat \neq output?

lukesonnet: Github also says that I dismissed your stale reviews. I have done no such thing.

R/S3_predict.R
```diff
@@ -74,13 +74,21 @@ predict.lm_robust <- function(object,
 
   # Get model matrix
   rhs_terms <- delete.response(object$terms)
+  if (object[["fes"]]) {
+    print(str(X))
+    print(rhs_terms)
+  }
```
nfultz: debugging code snuck through?

R/estimatr_iv_robust.R
```diff
@@ -50,6 +55,16 @@
 #' first-stage (endogenous) regressors. More notes on this can be found at
 #' \href{https://declaredesign.org/R/estimatr/articles/mathematical-notes.html}{the mathematical appendix}.
 #'
+#' If \code{`fixed_effects`} are specified, both the outcome, regressors, and instruments
```
nfultz: I don't think you need both code and backticks.

R/estimatr_iv_robust.R
```diff
@@ -129,6 +152,14 @@ iv_robust <- function(formula,
     warning("More regressors than instruments")
   }
 
+  fes <- !is.null(model_data[["fixed_effects"]])
```
nfultz: prefer is.formula over !is.null

R/helper_clean_model_data.R
```diff
@@ -129,3 +153,35 @@ clean_model_data <- function(data, datargs, estimator = "") {
 
   return(ret)
 }
+
+demean_fes <- function(model_data) {
+  nfaclevels <-
+    apply(model_data[["fixed_effects"]], 2, function(fe) length(unique((fe)))-1)
```
nfultz: are these already factors / ints?

```
vapply(model_data$fixed_effects, nlevels, 1)
```

R/estimatr_iv_robust.R
lukesonnet: It's not a formula, and it's probably a character, but it could be numeric. I am going to stick with the `!is.null` on this one :/

R/helper_clean_model_data.R
lukesonnet: This returns a matrix if there are multiple fixed effects.

.travis.yml
```diff
@@ -24,7 +24,7 @@ matrix:
 
     - os: osx
       osx_image: xcode6.4
-      r: oldrel
+      r: 3.4
```
nfultz: also get rid of xcode6.4 setting

.travis.yml
```diff
@@ -44,7 +44,8 @@ addons:
 r_github_packages:
 - DeclareDesign/DDtools
 - ropensci/git2r
-- DeclareDesign/fabricatr@v0.4.0
+- DeclareDesign/fabricatr
+- DeclareDesign/randomizr
```
nfultz: Has something changed in randomizr that you need?

R/S3_predict.R
```diff
@@ -184,20 +184,75 @@ predict.iv_robust <- function(object,
...
+  femat <- model.matrix(
+    ~ 0 + .,
+    data = as.data.frame(
+      sapply(
+        stats::model.frame.default(
+          args[["fixed_effects"]],
+          data = newdata,
+          na.action = NULL
+        ),
+        function(fe) as.factor(fe)
```
nfultz: as.factor

nfultz: consider writing a helper function, I've seen this block before.

nfultz: also use lapply instead of sapply, you don't want an extra simplify called accidentally.

R/S3_predict.R
```diff
@@ -184,20 +184,75 @@ predict.iv_robust <- function(object,
...
+      )
+    )
+  )
+
+  keep_facs <- names(object[["fixed_effects"]]) %in% colnames(femat)
+  extra_facs <- length(setdiff(colnames(femat), names(object[["fixed_effects"]])))
```
nfultz: Was the intent to print the levels in the error message?

If not rewrite to `!all(colnames(femat) %in% names(...))`

Also use setdiff for keep_facs so you don't have to repeat the `names(object[[...` crud below.

R/S3_print.R
```diff
@@ -34,9 +34,7 @@ print_summary_lm_like <- function(x,
   print(coef(x), digits = digits)
 
   if (!is.null(x$fstatistic)) {
-    cat(
-      "\nMultiple R-squared: ", formatC(x$r.squared, digits = digits),
-      ",\tAdjusted R-squared: ", formatC(x$adj.r.squared, digits = digits),
+    fstat <- paste(
```
nfultz: is.numeric

also use a ternary if

R/helper_clean_model_data.R
nfultz: use args_ignored as above.

R/helper_clean_model_data.R
```diff
@@ -119,13 +138,57 @@ clean_model_data <- function(data, datargs, estimator = "") {
...
       "`condition_prs` must be a vector of positive values no greater than 1"
     )
   }
 
   ret[["terms"]] <- attr(mf, "terms")
+  ret[["xlevels"]] <- .getXlevels(ret[["terms"]], mf)
+  if (!is.null(ret[["fixed_effects"]])) {
+    ret[["felevels"]] <-
+      setNames(
+        lapply(colnames(ret[["fixed_effects"]]), function(x) {
```
nfultz: ```lapply(as.data.frame(ret[["fixed_effects"]]), unique) ```

R/helper_return.R
```diff
@@ -1,9 +1,11 @@
 # This file has helper functions for returning the lists from various estimators
-lm_return <- function(return_list, model_data, formula) {
+lm_return <- function(return_list, model_data, formula, fes = FALSE) {
```
nfultz: is fes used anywhere

src/lm_robust_helper.cpp
```diff
@@ -163,34 +289,70 @@ List lm_variance(const Eigen::Map<Eigen::MatrixXd>& X,
...
+    if ( !clustered ) {
+      // Rcout << "temp_omega:" << std::endl << temp_omega << std::endl;
 
-      } else if (type == "HC3") {
+      if ((se_type == "HC2") || (se_type == "HC3")) {
 
         for (int i = 0; i < n; i++) {
-          Eigen::VectorXd Xi = X.row(i);
-          temp_omega.row(i) = temp_omega.row(i) / (std::pow(1.0 - Xi.transpose() * XtX_inv * Xi, 2));
+          Eigen::ArrayXd new_omega(ny);
```
nfultz: Minor nitpick don't call the constructor if you are going to reassign it below anyway. Declare outside the loop as a pointer.

src/lm_robust_helper.cpp
```diff
@@ -163,34 +289,70 @@ List lm_variance(const Eigen::Map<Eigen::MatrixXd>& X,
...
+            new_omega = temp_omega.row(i) / (std::pow(1.0 - Xi.transpose() * meatXtX_inv * Xi, 2));
+          }
+          // Rcout << i << ":" << new_omega << std::endl;
+          // Perfect fits cause instability, but we can place 0s for those
+          // observations and the rest of the estimation works
+          for (int j = 0; j < ny; j++) {
+            if (std::isnan(new_omega(j)) || std::isinf(new_omega(j))) {
+              new_omega(j) = 0.0;
+            }
+          }
```
nfultz: https://stackoverflow.com/questions/22927323/how-do-i-find-and-replace-all-non-finite-numbers-in-an-eigenarray-object

R/helper_clean_model_data.R
```diff
@@ -119,13 +138,57 @@ clean_model_data <- function(data, datargs, estimator = "") {
...
+    weights = if (is.numeric(model_data[["weights"]])) model_data[["weights"]] else rep(1, nrow(model_data[["design_matrix"]])),
+    has_int = attr(model_data$terms, "intercept"),
+    eps = 1e-8
+  )
+
+  # save names
+  dimnames(demeaned[["outcome"]]) <- dimnames(model_data[["outcome"]])
+  new_names <- dimnames(model_data[["design_matrix"]])
+  new_names[[2]] <- new_names[[2]][new_names[[2]] != "(Intercept)"]
+  dimnames(demeaned[["design_matrix"]]) <- new_names
```
nfultz: set the names etc inside demeanMat

R/helper_clean_model_data.R
lukesonnet: I have to rewrite this a bit, because we actually don't want to ignore "subset" (and we don't actually ignore "args_ignored" right now).

R/S3_predict.R
lukesonnet: Yes you've seen it before, but the other one is evaluating a quosure inside the apply.

R/S3_predict.R
lukesonnet: > Also use setdiff for keep_facs so you don't have to repeat the names(object[[... crud below.

Do you mean intersect?

src/lm_robust_helper.cpp
lukesonnet: Thx

src/lm_robust_helper.cpp
lukesonnet: Blessed.

R/helper_clean_model_data.R
lukesonnet: Eigen matrices don't have names?

lukesonnet:
Sorry, one last pass @nfultz (merging is blocked without reviews)

PR #220, opened 2018-05-31
Miscellaneous 0.8 patches

lukesonnet:


R/S3_vcov.R
```diff
@@ -1,38 +1,45 @@
 #' @export
 vcov.lm_robust <-
-  function(object, ...) {
-    return(vcov_simple(object))
+  function(object, complete = TRUE, ...) {
+    return(vcov_simple(object, complete = complete))
```
nfultz: use implicit returns, function goes on same line as variable assignment.

PR #228, opened 2018-06-08
Add statistic to output

lukesonnet:


R/helper_starprep.R
```diff
@@ -197,7 +197,7 @@ commarobust <- function(model,
 #' Prepare model fits for stargazer
 #'
 #' @param ... a list of lm_robust or lm objects
-#' @param stat either "std.error" (the default), "p.value", "ci", or "df"
+#' @param stat either "std.error" (the default), "statistic" (the t-statistic), "p.value", "ci", or "df"
```
graemeblair: should this be the t-statistic, or should it refer to either Z or t statistic?

lukesonnet: So (1) the name of the vector in the `lm_robust` object is called `statistic` because we go for `tidy` names whenever there isn't a pre-existing `lm` name, and (2) so as with the other options for `stat` except for `ci`, the argument is the same as the name in the `lm_robust` output.

We do have z-stats in HT, but `commarobust` only works on `lm` objects and is always t-stats. So we could also have tstatistic or some variant work if you want.

graemeblair: I just meant in the help - it seemed like sometimes starprep would get z and sometimes t

lukesonnet: It will never get a z-stat because only HT has z-stats.

nfultz:
You should think about using roxygen @inheritparams, this PR is small but touched many files just to update docs

lukesonnet:
Was just thinking about that @nfultz. Thanks for the heads up.

PR #241, opened 2018-07-03
Add support for one group in a fixed effect

lukesonnet:
Solves Macartan's problem about wanting support for their being a fixed effect with only one group. Note for now we do not support a fixed effect with only one group if there are multiple fixed effect variables.

Also solves a bug where if there is one fixed effect variable and missingness, `clean_model_data` returns a vector rather than a matrix for the fixed effects.

### DIFF
--- a/R/helper_clean_model_data.R
+++ b/R/helper_clean_model_data.R
@@ -142,6 +142,11 @@ clean_model_data <- function(data, datargs, estimator = "") {
   ret[["condition_pr"]] <- model.extract(mf, "condition_pr")
 
   ret[["fixed_effects"]] <- model.extract(mf, "fixed_effects")
+  # If there is NA in the blocks and only one block, returns vector not matrix
+  # so coerce to matrix
+  if (is.character(ret[["fixed_effects"]])) {
+    ret[["fixed_effects"]] <- as.matrix(ret[["fixed_effects"]])
+  }
 
   if (any(ret[["condition_pr"]] <= 0 | ret[["condition_pr"]] > 1)) {
     stop(
--- a/R/helper_lm_robust_fit.R
+++ b/R/helper_lm_robust_fit.R
@@ -565,9 +565,27 @@ prep_data <- function(data,
   }
 
   if (fes) {
-    data[["femat"]] <- model.matrix(
-      ~ 0 + .,
-      data = as.data.frame(data[["fixed_effects"]])
+    data[["femat"]] <- tryCatch({
+      model.matrix(
+        ~ 0 + .,
+        data = as.data.frame(data[["fixed_effects"]])
+      )},
+      error = function(e) {
+        msg <- conditionMessage(e)
+        if (msg == "contrasts can be applied only to factors with 2 or more levels") {
+          if (ncol(data[["fixed_effects"]]) == 1) {
+            matrix(
+              rep(1, nrow(data[["fixed_effects"]])),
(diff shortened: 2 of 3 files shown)


R/helper_lm_robust_fit.R
```diff
@@ -565,10 +565,21 @@ prep_data <- function(data,
   }
 
   if (fes) {
-    data[["femat"]] <- model.matrix(
-      ~ 0 + .,
-      data = as.data.frame(data[["fixed_effects"]])
-    )
+    fe_dat <- as.data.frame(data[["fixed_effects"]])
+    fe_levels <- vapply(fe_dat, nlevels, numeric(1))
```
nfultz: I think you can use `0` or similar constant instead of `numeric(1)`

R/helper_lm_robust_fit.R
```diff
@@ -565,10 +565,21 @@ prep_data <- function(data,
...
+    if (any(fe_levels == 1)) {
+      if (ncol(fe_dat) == 1) {
+        data[["femat"]] <- matrix(
+          rep(1, nrow(data[["fixed_effects"]])),
+          dimnames = list(attr(data[["fixed_effects"]], "names"),
+                          paste0(colnames(data[["fixed_effects"]]), data[["fixed_effects"]][1]))
+        )
+      } else {
```
nfultz: Tighten the branch a little.

```
if(ncol(fe_data) != 1){
  stop(...)
}
data[["femat"]] <- ...
```

lukesonnet:
@nfultz @graemeblair need an updated review here.

R/helper_clean_model_data.R
```diff
@@ -142,6 +142,11 @@ clean_model_data <- function(data, datargs, estimator = "") {
   ret[["condition_pr"]] <- model.extract(mf, "condition_pr")
 
   ret[["fixed_effects"]] <- model.extract(mf, "fixed_effects")
+  # If there is NA in the blocks and only one block, returns vector not matrix
+  # so coerce to matrix
+  if (is.character(ret[["fixed_effects"]])) {
```
nfultz: to check if it's a matrix, usually you do `is.matrix` - since as.matrix is so cheap, may as well do it unconditionally though.

R/helper_lm_robust_fit.R
```diff
@@ -565,10 +565,23 @@ prep_data <- function(data,
...
+    fe_dat <- as.data.frame(data[["fixed_effects"]])
+    fe_levels <- vapply(fe_dat, nlevels, 0L)
+    if (any(fe_levels == 1)) {
+      if (ncol(fe_dat) != 1) {
+        stop(
+          "Can't have a fixed effect with only one group AND multiple fixed ",
+          "effect variables"
+        )
+      }
+      data[["femat"]] <- matrix(
```
nfultz: no rep reqd

```
matrix(1, nrow(fe_dat), ... 
```

R/helper_lm_robust_fit.R
```diff
@@ -565,10 +565,23 @@ prep_data <- function(data,
...
+        rep(1, nrow(data[["fixed_effects"]])),
+        dimnames = list(attr(data[["fixed_effects"]], "names"),
```
nfultz: use names() and fe_dat to be more legible

lukesonnet:
Travis bugged out, once I can restart that will merge, then will incorporate NF suggestions and CRAN prep in different PR so it reqs only one more approval/read through.

PR #243, opened 2018-07-11
Ls/cran

lukesonnet:
Makes some cosmetic code changes, allows HT to take a single value for condition_prs instead of just a variable name.

### DIFF
--- a/R/S3_summary.R
+++ b/R/S3_summary.R
@@ -102,8 +102,7 @@ summary.horvitz_thompson <- function(object, ...) {
 summarize_tidy <- function(object, test = "t", ...) {
   remove_cols <- c("term", "outcome")
 
-  # This is ugly SO THAT summary(fit)$coefficients returns
-  # something like lm does.
+  # Ugly so that summary(fit)$coefficients matches lm()
   tidy_out <- tidy(object, ...)
   colnames(tidy_out)[2:8] <-
     c(
--- a/R/S3_summary.R
+++ b/R/S3_summary.R
@@ -102,7 +102,8 @@ summary.horvitz_thompson <- function(object, ...) {
 summarize_tidy <- function(object, test = "t", ...) {
   remove_cols <- c("term", "outcome")
 
-  # This is ugly SO THAT summary(fit)$coefficients returns something like lm does.
+  # This is ugly SO THAT summary(fit)$coefficients returns
+  # something like lm does.
   tidy_out <- tidy(object, ...)
   colnames(tidy_out)[2:8] <-
     c(
--- a/R/estimatr_difference_in_means.R
+++ b/R/estimatr_difference_in_means.R
@@ -454,10 +454,8 @@ difference_in_means_internal <- function(condition1 = NULL,
   # Check to make sure multiple in each group if pair matched is false
   if (!pair_matched & (N2 == 1 | N1 == 1)) {
     stop(
-      "Must have least two treated/control units in each block if design is ",
-      "not pair-matched (i.e., every block is of size two). Only one treated ",
-      "or control unit in a block makes standard errors impossible to ",
-      "calculate"
+      "If design is not pair-matched, every block must have at least two ",
+      "treated and control units."
     )
   }
 
--- a/R/estimatr_difference_in_means.R
+++ b/R/estimatr_difference_in_means.R
@@ -288,7 +288,8 @@ difference_in_means <- function(formula,
   } else {
     pair_matched <- FALSE
 
-    # When learning whether it is matched pairs, should only use relevant conditions
+    # When learning whether design is matched pairs,
+    # should only use rows in relevant conditions
     data <- subset.data.frame(data, t %in% c(condition1, condition2))
 
     clust_per_block <- check_clusters_blocks(data)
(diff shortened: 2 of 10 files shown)


R/helper_clean_model_data.R
```diff
@@ -21,13 +21,16 @@ clean_model_data <- function(data, datargs, estimator = "") {
...
 
-  args_ignored <- c("fixed_effects", "se_type")
+  args_ignored <- c("fixed_effects", "condition_pr", "se_type")
   # For each ... that would go to model.fram .default, early eval,
   # save to formula env, and point to it
   # subset is also non-standard eval
   to_process <- setdiff(
     names(mfargs),
-    c(setdiff(names(formals(stats::model.frame.default)), "subset"), args_ignored)
+    c(
```
nfultz: this would be clearer as `setdiff(setdiff(...), args_ignored)` - add a small infix helper to utils.R `%setdiff%`

R/helper_clean_model_data.R
```diff
@@ -49,6 +52,19 @@ clean_model_data <- function(data, datargs, estimator = "") {
...
+  recycle_prs <- FALSE
+  if ("condition_pr" %in% names(mfargs)) {
+    condition_pr <- eval_tidy(mfargs[["condition_pr"]], data = data)
+    if (length(condition_pr) == 1) {
+      recycle_prs <- TRUE
+      mfargs[["condition_pr"]] <- NULL
+    } else {
+      name <- sprintf(".__condition_pr%%%d__", sample.int(.Machine$integer.max, 1))
+      m_formula_env[[name]] <- condition_pr
+      mfargs[["condition_pr"]] <- sym(name)
```
nfultz: Let the loop above do this part (else block), this block can clean it up if it's length == 1.

R/helper_lm_robust_fit.R
```diff
@@ -575,9 +577,12 @@ prep_data <- function(data,
         )
       }
       data[["femat"]] <- matrix(
-        rep(1, nrow(data[["fixed_effects"]])),
-        dimnames = list(attr(data[["fixed_effects"]], "names"),
-                        paste0(colnames(data[["fixed_effects"]]), data[["fixed_effects"]][1]))
+        1,
+        nrow(data[["fixed_effects"]]),
+        dimnames = list(
+          attr(data[["fixed_effects"]], "names"),
```
nfultz: use names() instead of attr(x, "names")

R/S3_summary.R
```diff
@@ -102,7 +102,8 @@ summary.horvitz_thompson <- function(object, ...) {
 summarize_tidy <- function(object, test = "t", ...) {
   remove_cols <- c("term", "outcome")
 
-  # This is ugly SO THAT summary(fit)$coefficients returns something like lm does.
+  # This is ugly SO THAT summary(fit)$coefficients returns
+  # something like lm does.
```
nfultz: Edit the comment instead of wrapping it, eg

```
  # Ugly so that summary(fit)$coefficients matches lm() 
```

R/estimatr_difference_in_means.R
```diff
@@ -447,12 +451,13 @@ difference_in_means_internal <- function(condition1 = NULL,
...
+  # Check to make sure multiple in each group if pair matched is false
   if (!pair_matched & (N2 == 1 | N1 == 1)) {
     stop(
-      "Must have least two treated/control units in each block if design is not ",
-      "pair-matched (i.e., every block is of size two). Only one treated or ",
-      "control unit in a block makes standard errors impossible to calculate"
+      "Must have least two treated/control units in each block if design is ",
+      "not pair-matched (i.e., every block is of size two). Only one treated ",
+      "or control unit in a block makes standard errors impossible to ",
+      "calculate"
```
nfultz: ``` Std errors are undefined when only one unit is in a block```

R/helper_clean_model_data.R
nfultz: of `%s-%` for short.

R/helper_clean_model_data.R
```diff
@@ -139,7 +155,10 @@ clean_model_data <- function(data, datargs, estimator = "") {
 
   ret[["block"]] <- model.extract(mf, "block")
 
-  ret[["condition_pr"]] <- model.extract(mf, "condition_pr")
+  ret[["condition_pr"]] <- if (recycle_prs)
+    rep(condition_pr, nrow(ret[["design_matrix"]]))
```
nfultz: use rep_len, then you don't need the `recycle_prs` flag at all.

R/estimatr_difference_in_means.R
lukesonnet: That's not correct but I'll shorten it.

R/helper_clean_model_data.R
lukesonnet: not right now thanks.

R/helper_clean_model_data.R
lukesonnet: will fix

R/helper_lm_robust_fit.R
lukesonnet: will fix

R/S3_summary.R
lukesonnet: will fix

R/helper_clean_model_data.R
lukesonnet: They come from different sources, so I think I still need it.

lukesonnet:
Need reviews again

lukesonnet:
Once this passes I will merge and immediately submit to CRAN. I'm not doing valgrid checks etc... because nothing too major has changed, no C++ has changed, probably the only substantial change is some tweaks in clean_model_data and handling the case of fixed_effects with one block.

PR #256, opened 2018-09-01
Fixes two bugs with multivariate regression

lukesonnet:
Fixes bug that truncated degrees of freedom with "CR2"  standard errors and only returned them for first outcome

Fixes bug that returned incorrect R-squared for the second or later outcomes


lukesonnet:
Unless this goes straight to CRAN it shouldn't be 0.12.0.

It should probably be 0.11.1 or 0.11.2. I'll pay more attention to this in the future as there are users now!

PR #262, opened 2018-10-19
Move `tidy` to `broom`

lukesonnet:
This PR removes all `tidy` functions from `estimatr` and instead DEPENDS on `broom`.

I had to re-rig a bunch of the tests because of problems with `tibble` and checking for equivalence.

This PR means that we should not resubmit to CRAN until the next version of broom goes on with the estimatr tidiers.


tests/testthat/test-difference-in-means.R
```diff
@@ -110,20 +110,20 @@ test_that("DIM Blocked", {
   expect_equivalent(dim_01$coefficients, 4)
 
   expect_equivalent(
-    tidy(dim_01),
-    tidy(
+    as.data.frame(tidy(dim_01)),
```
nfultz:
```
expect_equiv_tbl <- function(e1, e2) expect_equivalent(as.data.frame(e1), as.data.frame(e2))

File: DESCRIPTION [hunk 2]
```diff
@@ -17,10 +17,11 @@ Depends: R (>= 3.4.0)
 Imports:
     Rcpp (>= 0.12.16),
     methods,
+    broom,
```
nfultz: Is tidy ever used in the actual package? If only used in tests, can make broom a suggests.

lukesonnet:
I think this means I have to change the branch that is PRd here, no? https://github.com/tidymodels/broom/pull/511

As I need `master` to be local, temporary version for drat

DESCRIPTION
lukesonnet: Yes it is, in a couple of places (e.g. `commarobust`).

lukesonnet:
We're changing strategies to rely on `generics`.

Issue #261, opened 2018-10-16
Add CR3 clustered se option


acoppock:

lukesonnet:
This is finished, but I'm refactoring some of the testing code before I PR it.

PR #263, opened 2018-10-23
Adds CR3 estimator to lm_robust and iv_robust

lukesonnet:
- Version bump for first post-0.12 release bump
- Adds CR3 standard error estimators to both `lm_robust` and `iv_robust` (Issue #261)
- Updates documentation and tests (lots of test refactoring)
- For now gives S-1 as the degrees of freedom, as the next PR will just divide the DoF from the SE

### DIFF
--- a/src/lm_robust_helper.cpp
+++ b/src/lm_robust_helper.cpp
@@ -431,31 +431,18 @@ List lm_variance(Eigen::Map<Eigen::MatrixXd>& X,
 
           } else if (se_type == "CR3") {
 
-            //if (Xunweighted.isNotNull()) {
+            Eigen::CompleteOrthogonalDecomposition<Eigen::MatrixXd> At(
+                Eigen::MatrixXd::Identity(len, len) -
+                  Xoriginal.block(start_pos, 0, len, r_fe) *
+                  meatXtX_inv *
+                  X.block(start_pos, 0, len, r_fe).transpose()
+            );
 
-              Eigen::CompleteOrthogonalDecomposition<Eigen::MatrixXd> At(
-                  Eigen::MatrixXd::Identity(len, len) -
-                    Xoriginal.block(start_pos, 0, len, r_fe) *
-                    meatXtX_inv *
-                    X.block(start_pos, 0, len, r_fe).transpose()
-              );
-
-            //}
-
-            // Eigen::ColPivHouseholderQR<Eigen::MatrixXd> At(
-            //     Eigen::MatrixXd::Identity(len, len) -
-            //       Xoriginal.block(start_pos, 0, len, r_fe) *
-            //       meatXtX_inv *
-            //       X.block(start_pos, 0, len, r_fe).transpose()
-            // );
             if (Xunweighted.isNotNull()) {
               At_WX_inv =  At.solve(Eigen::MatrixXd::Identity(len, len)).transpose() * X.block(start_pos, 0, len, r_fe);
             } else {
-              // Could speed up using LLT instead of QR for unweighted case as At is symmetric in that case
               At_WX_inv =  At.solve(X.block(start_pos, 0, len, r_fe));
             }
-
-            // Rcout << "At_WX_inv: " << std::endl << At_WX_inv << std::endl;
 
           }
 
--- a/src/lm_robust_helper.cpp
+++ b/src/lm_robust_helper.cpp
@@ -357,6 +357,7 @@ List lm_variance(Eigen::Map<Eigen::MatrixXd>& X,
       }
 
       Eigen::Map<Eigen::ArrayXi> clusters = Rcpp::as<Eigen::Map<Eigen::ArrayXi> >(cluster);
+      // Rcout << "clusters: "<< std::endl << clusters << std::endl;
 
       double current_cluster = clusters(0);
       int clust_num = 0;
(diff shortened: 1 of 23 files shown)


tests/testthat/helper-return-cleaners.R
```diff
@@ -0,0 +1,13 @@
+# This fn removes calls from function returns to make testing easier
+rmcall <- function(obj) {
+  if (!is.null(obj[["call"]])) {
+    obj[["call"]] <- NULL
+  }
+  return(obj)
```
nfultz:
```
rmcall <- function(obj) { structure(obj, call=NULL) }
```

src/lm_robust_helper.cpp
```diff
@@ -426,16 +428,30 @@ List lm_variance(Eigen::Map<Eigen::MatrixXd>& X,
...
+
+            Eigen::CompleteOrthogonalDecomposition<Eigen::MatrixXd> At(
+                Eigen::MatrixXd::Identity(len, len) -
+                  Xoriginal.block(start_pos, 0, len, r_fe) *
+                  meatXtX_inv *
+                  X.block(start_pos, 0, len, r_fe).transpose()
+            );
+
+            if (Xunweighted.isNotNull()) {
+              At_WX_inv =  At.solve(Eigen::MatrixXd::Identity(len, len)).transpose() * X.block(start_pos, 0, len, r_fe);
```
nfultz: Why solving against Identity? Maybe add a comment.

src/lm_robust_helper.cpp
```diff
@@ -447,7 +463,7 @@ List lm_variance(Eigen::Map<Eigen::MatrixXd>& X,
 
           } else {
 
-            if (se_type == "CR2") {
+            if (se_type == "CR2" || se_type == "CR3") {
```
nfultz: Choose if you want 
 
    ((se_type == "CR2") || (se_type == "CR3"))

    (se_type == "CR2" || se_type == "CR3")

    (se_type %in% c("CR2", "CR3"))

and be consistent throughout.

tests/testthat/test-iv-robust-fes.R
```diff
@@ -16,313 +16,92 @@ dat$Xdup <- dat$X
...
-  #   tidy(ro)[ro$term %in% c("X1", "X2"), ],
-  #   tidy(rfo)[rfo$term %in% c("X1", "X2"), ]
-  # )
-  #
-  # expect_equal(
-  #   ro[c("r.squared", "adj.r.squared")],
-  #   rfo[c("r.squared", "adj.r.squared")]
-  # )
 
+  for (se_type in se_types) {
```
nfultz: +1

tests/testthat/test-iv-robust.R
```diff
@@ -10,6 +10,7 @@ dat <- data.frame(
   clust = sample(letters[1:3], size = N, replace = TRUE)
 )
 dat$z <- dat$x * 0.5 + rnorm(N)
+dat$x1_c <- dat$x
```
nfultz: should it be centered?

nfultz: nm, maybe c is for copy

src/lm_robust_helper.cpp
lukesonnet: In both case we need the transpose, but unweighted case is symmetric so it doesn't matter. I could instead transpose the matrix before hand (as you can't transpose the `CompleteOrthogonalDecomposition`).

tests/testthat/helper-return-cleaners.R
nfultz: ah - right, different from DeclareDesign. You can still remove the null check I think.

src/lm_robust_helper.cpp
nfultz: If Eigen knows the solution is sym, the transpose should be a no-op - I generally prefer the more concise and general expression over the two branches that are slightly different. However, there's a numerical and performance penalty to creating the inverse explicitly when it isn't actually needed. 


Maybe you can move the transpose inside a la https://math.stackexchange.com/a/340234/139343

lukesonnet: It doesn't know the solution is symmetric, so the transpose is not a no-op, right?

nfultz: It /should/ know if an object is symmetric or not. When inner products are used to calculate eg covariance matrices, the result is  supposed to be subclassed as self-adjoint and further operators dispatched to symmetric implementations. It's an easy 50% speed up for many problems (eg crossprod(X) vs t(X) %*% X, chol(X) vs qr(X), etc). 

Even if you store a symmetric matrix in a regular Matrix variable, it's methods will still be dispatched to symmetric functions appropriately, which can then also return symmetric matrices will their own appropriate methods (this is `covariance` in the CS parlance https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science) ).

See the RcppEigen Vignette section on Crossproduct of a matrix -  IIRC that's roughly where you had started in the first place, but I might be thinking of Armadillo. There's also a chance I'm getting the two libraries confused.

macartan:
Luke -- has it be possible to check this implementation against others in R
or stata?

src/lm_robust_helper.cpp
nfultz: also transposes are basically free compared to everything else you're doing.

lukesonnet:
@macartan, the tests compare it with clubSandwich. Stata does not have CR3 standard errors, but I think base `sandwich` in R does now and I could add those tests.

lukesonnet:
I'm stuck right now because the slight (numerical) differences in the variance-covariance matrix from commarobust or from lm_robust are causing different kinds of failures in computing the F-statistic on windows 32bit machines.

This is an absolutely ridiculous error, but I'm trying to eliminate or minimize the numerical deviations that are causing it.

acoppock:
OK with me if we just round to the 10th decimal place or something to get around this windows issue

acoppock:
Thanks so much for adding this.

PR #265, opened 2018-11-04
Add .onLoad warning for old versions of broom

lukesonnet:
This PR will `packageStartupMessage` if `broom` version <= 0.5.0, the current CRAN version, is attached. If they attach broom afterwards, there's no guarantee.

Hopefully the `broom` that relies on `generics` will be on CRAN soon.

This also tests the message is silent if the right version of `broom` is installed and that it messages with the old version. Travis hopefully is set up to test both on Linux, release R.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ matrix:
   include:
     - os: linux
       r: release
+      r_github_packages: tidymodels/broom
+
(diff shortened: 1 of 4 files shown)


lukesonnet:
Also looks like Travis fixed their Fortran problem!

PR #267, opened 2018-11-05
Ignore broom test on CRAN

lukesonnet:


lukesonnet:
OK this passes on Travis w/ warning and rhub w/o warning (as that only gets the built tests). Please approve and will resubmit.

PR #277, opened 2018-12-13
Hyperlink DOIs to preferred resolver

katrinleinweber:
Hello :-)

The DOI foundation recommends [this new resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). It may be a bit ironic that they would change the URL of their service, but it's now [encrypted](https://www.ssllabs.com/ssltest/analyze.html?d=doi.org). Because the old `dx` subdomain is being redirected, there is no urgent need to do anything.

However, I'd hereby like to suggest to follow the new recommendation by applying this update to all static DOI links.

Cheers!

### DIFF
--- a/R/estimatr_iv_robust.R
+++ b/R/estimatr_iv_robust.R
@@ -100,7 +100,7 @@
 #'
 #' @references
 #'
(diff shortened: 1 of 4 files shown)


graemeblair:
@katrinleinweber thanks so much for this PR & good to know about the new resolver!

katrinleinweber:
You're welcome, and thanks for merging :-)

lukesonnet:
Awesome, really appreciate it.

PR #281, opened 2019-01-19
combine coefficients

graemeblair:
Just opening a PR so we can track this work


graemeblair:
This really works nicely, thank you for this work.

some edits:

1.  The `lh_label` we think can be removed; seems reasonable that the estimates be named after their linear hypothesis text. are there any cases where this would break?  We tested some multiple test things and it appeared to work fine, e.g.:

```{r}
library(DeclareDesign)

design <- 
  declare_population(N = 100, u = rnorm(N), X = rbinom(N, 1, 0.5)) +
  declare_potential_outcomes(Y ~ 3 * Z + 2 * Z * X + u) +
  declare_estimand(cate = mean(Y_Z_1 - Y_Z_0), subset = X == 1) +
  declare_assignment(m = 50) +
  declare_reveal(Y, Z)

dat <- draw_data(design)

fit_1 <- lh_robust(Y ~ Z*X, data = dat, linearHypothesis = "Z + Z:X = 0")
fit_2 <- lh_robust(Y ~ Z*X, data = dat, linearHypothesis = c("(Intercept) = 1", "Z = 1"))

tidy(fit_1)
tidy(fit_2)
```


2. Could you remove the rownames from the tidy dataset? I don't think that estimatr's methods ever return dfs with rownames.

3. We noticed an "lh" class -- what's that about?

4. Could you write tests that confirm we get things right regardless of the standard error estimator?  here's a little start (not sure how to get coefficients out of car to compare)

```{r}
fit_1 <- lh_robust(Y ~ Z*X, data = dat, linearHypothesis = "Z + Z:X = 0")
fit_lmr <- lm_robust(Y ~ Z*X, data = dat)
car_lmr <- car::linearHypothesis(fit_2, hypothesis.matrix = "Z + Z:X = 0")

fit_lm <- lm_robust(Y ~ Z*X, data = dat)
car_lm <- linearHypothesis(fit_2, hypothesis.matrix = "Z + Z:X = 0", white.adjust = "HC2")

summary(fit_1)
car_lmr
car_lm
```
5. Could you write some examples in the documentation for the lh_robust function?

Thank you thank you.

R/S3_summary.R
```diff
@@ -130,3 +130,18 @@ summarize_tidy <- function(object, test = "t", ...) {
 
   return(tidy_mat)
 }
+
+#' @export
+summary.lh <- function(x,...){
+  print(x)
+  cat("\n")
+  print(x$linearHypothesis)
```
nfultz: summary should not print anything itself. It should return an object to be printed by the REPL. eg `stats::summary.lm` and `stats:::print.summary.lm`

R/S3_tidy.R
```diff
@@ -81,3 +81,21 @@ tidy.difference_in_means <- tidy_data_frame
...
 tidy.horvitz_thompson <- tidy_data_frame
+
+
+#' @rdname estimatr_tidiers
+#' @templateVar class lh_wrapper
+#'
+#' @export
+#' @family estimatr tidiers
+tidy.lh_robust <- function(x, ...) {
+   do.call(rbind, lapply(x, tidy_data_frame))
```
nfultz: Overly general, will break if the object gets modified and holds a non-tidyable thing. Instead prefer

```
rbind(tidy(x$lm_robust), tidy(x$linear_hypothesis))
```
which is specific and has a clear intent.

R/estimatr_lh_robust.R
```diff
@@ -0,0 +1,161 @@
...
+
+
+
+  attr( return_lh , "class") <- "lh"
+
+
+  return_list <- list(lm_robust =  return_lm_robust,
+                      linearHypothesis =  return_lh )
+
+  attr(   return_list , "class") <- "lh_robust"
```
nfultz:
```suggest
  class(return_list) <- "lh_robust"
```

R/estimatr_lh_robust.R
```diff
@@ -0,0 +1,161 @@
...
+  std.error <- sqrt(diag(attr(out, "vcov")))
+  df <- model$df.residual
+  tt <- estimate/std.error
+  p.value <-  2 * pt(abs(tt), df)
+  alpha_ <- alpha/2
+  alpha_ <- c(alpha_, 1 - alpha_)
+  ci <- estimate + std.error %o% qt(alpha_, df)
+  if(is.null(lh_term)) lh_term <- linearHypothesis
+
+  return_lh <- list(   coefficients =  estimate,
```
nfultz: Why a list here and not a data.frame ?

R/estimatr_lh_robust.R
```diff
@@ -0,0 +1,161 @@
...
+                      return_vcov = TRUE,
+                      try_cholesky = FALSE,
+                      linearHypothesis = NULL,
+                      lh_term = NULL
+                      ) {
+
+  return_list <- list()
+
+
+  datargs <- enquos(
```
nfultz: instead of copy-pasting all of lm_robust here, just call it.

R/estimatr_lh_robust.R
```diff
@@ -66,60 +93,17 @@ lh_robust <- function(formula,
...
                       try_cholesky = FALSE,
-                      linearHypothesis = NULL,
-                      lh_term = NULL
+                      linearHypothesis = NULL
                       ) {
+    call <- match.call()
+    args <- as.list(call)
+    args[[1]] <- NULL
+    args$linearHypothesis <- NULL
+    model <- do.call(lm_robust, args)
```
nfultz: This is better, but check that NSE still works with weights, subset and add a test.

R/S3_print.R
```diff
@@ -132,6 +133,11 @@ print.summary.lh_robust <- function(x,...){
   print(summary(x$lm_robust))
   cat("\n\n$linearHypothesis \n \n")
   x <- x[[2]]
-  print(x$linearHypothesis)
+
+  print(attr(x, "linearHypothesis"))
 }
+#' @export
+print.summary.lh <- function(x, ...){
```
nfultz: If you are just calling the default method, you don't even need this specific generic.

R/S3_summary.R
medinali: sounds good

R/S3_tidy.R
medinali: implemented!

R/estimatr_lh_robust.R
medinali: done!

R/estimatr_lh_robust.R
medinali: output didn't fit in a data.frame due to linear hypothesis object. Now it's is a returned as an attribute.

R/estimatr_lh_robust.R
medinali: couldn't avoid this.

R/estimatr_lh_robust.R
medinali: This does fine with weights and subsets locally. Though it's not stable for tests and checks. Data and other arguments aren't found in the environment.

Copy-pasting lm_robust unless there are other suggestions

R/estimatr_lh_robust.R
```diff
@@ -0,0 +1,184 @@
...
+#' intervals, TRUE by default.
+#' @param alpha The significance level, 0.05 by default.
+#' @param return_vcov logical. Whether to return the variance-covariance
+#' matrix for later usage, TRUE by default.
+#' @param try_cholesky logical. Whether to try using a Cholesky
+#' decomposition to solve least squares instead of a QR decomposition,
+#' FALSE by default. Using a Cholesky decomposition may result in speed gains, but should only
+#' be used if users are sure their model is full-rank (i.e., there is no
+#' perfect multi-collinearity)
+#' @param linearHypothesis A character string or a matrix specifying combination.
```
lukesonnet: Maybe this is a good time to have `lm_lin` and `lh_robust` borrow argument names from `lm_robust` rather than repeating them? Would be good to just have them pull from the same set once for overlapping arguments.

R/estimatr_lh_robust.R
```diff
@@ -0,0 +1,184 @@
...
+#' dat <- fabricate(
+#'   N = 40,
+#'   y = rpois(N, lambda = 4),
+#'   x = rnorm(N),
+#'   z = rbinom(N, 1, prob = 0.4),
+#'   clusterID = sample(1:4, 40, replace = TRUE)
+#' )
+#'
+#' # Default variance estimator is HC2 robust standard errors
+#' lhro <- lh_robust(y ~ x + z, data = dat, linearHypothesis = "z + 2x = 0")
```
lukesonnet: The example and example comment shouldn't focus on the SE estimtor but rather the specification of the linearHypothesis, and it should show several different possible linearHypotheses.

R/estimatr_lh_robust.R
```diff
@@ -0,0 +1,184 @@
...
+                      weights,
+                      subset,
+                      clusters,
+                      fixed_effects,
+                      se_type = NULL,
+                      ci = TRUE,
+                      alpha = .05,
+                      return_vcov = TRUE,
+                      try_cholesky = FALSE,
+                      linearHypothesis = NULL
```
lukesonnet: I understand why we call it `linearHypothesis`, but to fit our style it should be `linear_hypothesis`, and I would put it just after fixed_effects given that it's the main reason people would use `lh_robust`. In fact, should we even let people use `lh_robust` without `linear_hypothesis` specified?

tests/testthat/helper-se-types.R
```diff
@@ -1,4 +1,4 @@
 # se_types for various files
 
-se_types <- c("classical", "HC0", "HC1", "HC2", "HC3")
+se_types <- c("classical" , "HC0", "HC1", "HC2", "HC3")
```
lukesonnet: Undo this whitespace change

R/estimatr_lh_robust.R
lukesonnet: Why can't we just call lm_robust and then have lh_robust add to the output?

R/estimatr_lh_robust.R
```diff
@@ -0,0 +1,184 @@
...
+  if(is.null(linearHypothesis)){
+    warning("No linear hypothesis test performed")
+    return( return_list)}
+
+  car_lht <- car::linearHypothesis(return_list, linearHypothesis,
+                               level = 1-alpha)
+
+  estimate  <- drop(attr(car_lht, "value"))
+  std.error <- sqrt(diag(attr(car_lht, "vcov")))
+  df <- return_list$df.residual
```
lukesonnet: Are these the right degrees of freedom for tests with, say, the CR2, satterthwaite-like degrees of freedom?

R/estimatr_lh_robust.R
lukesonnet: Ah I see this comment now. I don't know how to solve this problem.

nfultz: > Data and other arguments aren't found in the environment.

That sounds like you need to adjust the envir / overscoping for eval.

nfultz: Also, an alternative solution would be to use `...` instead of manually capturing arguments.

R/estimatr_lh_robust.R
medinali: Thanks for pointing at this. I certainly need to look a bit more carefully into it.

Though my sense would be that since this is a post-estimation analysis the degree of freedom should be the same as those used by the model.

R/estimatr_lh_robust.R
graemeblair: agreed, it should fail w/o a `linear_hypothesis`

lukesonnet:
Also @medinali, please add yourself as a contributor to the description file.

R/estimatr_lh_robust.R
nfultz: +1 for inheritParams - see eg https://github.com/DeclareDesign/DeclareDesign/blob/master/R/aaa.R#L199-L206 for how I did that in DeclareDesign

tests/testthat/helper-se-types.R
medinali: done!

R/estimatr_lh_robust.R
medinali: sounds good! just renamed it!

R/estimatr_lh_robust.R
medinali: ok makes sense. Kept both

R/estimatr_lh_robust.R
medinali: the `...` approach works great [see commit](https://github.com/DeclareDesign/estimatr/pull/281/commits/c04a5e53562d0af9d5cf258c0709eeb79a8f4f22).

 I was sure I had tried that already.

R/estimatr_lh_robust.R
medinali: Not doing that exactly.
Adding `...` arg and linking to `lm_robust`

R/S3_summary.R
```diff
@@ -130,3 +130,17 @@ summarize_tidy <- function(object, test = "t", ...) {
...
+
+#' @export
+summary.lh_robust <- function(object,...){
+  class(object) <- "summary.lh_robust"
+  object
+}
+
+#' @export
+summary.lh <- function(object,...){
+  attr(object, "linearHypothesis")
```
lukesonnet: 1 - shouldn't it be `linear_hypothesis`?
2 - if it should, then this isn't being tested, correct?

R/estimatr_lh_robust.R
```diff
@@ -0,0 +1,114 @@
...
+#'
+#' # Default variance estimator is HC2 robust standard errors
+#' lhro <- lh_robust(y ~ x + z, data = dat, linear_hypothesis = "z + 2x = 0")
+#'
+#' # The linear hypothesis argument can be specified as following
+#' lh_robust(y ~ x + z, data = dat, linear_hypothesis = "z = 2x")
+#' lh_robust(y ~ x + z, data = dat, linear_hypothesis = c("z = 1", "x = 2"))
+#' lh_robust(y ~ x + z, data = dat, linear_hypothesis = "2*x +1*z")
+#'
+#' # which are alle equivalent to
```
lukesonnet: Typo, consider just having them in one block and say they are all equivalent.

R/estimatr_lh_robust.R
```diff
@@ -0,0 +1,114 @@
...
+#' # Can use summary() to get more statistics on the main output and subcomponents.
+#' summary(lhro)
+#' summary(lhro$lm_robust)
+#' summary(lhro$lh)
+#'
+#' @importFrom car linearHypothesis
+#'
+#' @export
+#'
+lh_robust <- function(..., data, linear_hypothesis) {
```
lukesonnet: This is much better! Awesome.

tests/testthat/test-lh-robust.R
```diff
@@ -0,0 +1,119 @@
...
+N <- 40
+dat <- data.frame(
+  Y = rnorm(N),
+  Y2 = rnorm(N),
+  Z = rbinom(N, 1, .5),
+  X = rnorm(N),
+  B = factor(rep(1:2, times = c(8, 12))),
+  cl = sample(1:4, size = N, replace = T),
+  w = runif(N)
+)
```
lukesonnet: You don't need to have perfect cluster or blocked variables (as is necessary in `difference_in_means`). I would just use `mtcars` for all of these tests.

lukesonnet:
Agree 2 Neal re: `car`, it imports a lot of junk we don't need at the moment.

R/S3_summary.R
graemeblair: yes fixing thanks

R/estimatr_lh_robust.R
graemeblair: thx

graemeblair:
thanks, comments addressed!

R/S3_summary.R
lukesonnet: bug remains. pls add tests. also what returns objects of class "lh"

lukesonnet: Oh I think I misunderstand. is `lh` the class of object linearHypothesis returns?

Issue #282, opened 2019-01-22
add a "glance" method


acoppock:
Someone made a new regression output package built on the soon-to-be-released `gt` package. Estimatr isn't compatbile yet because we need a "glance" method:

> library(gtsummary)
> library(estimatr)
> 
> fit_1 <- lm_robust(mpg ~ hp, data = mtcars)
> fit_2 <- lm_robust(mpg ~ hp, se_type = "classical", data = mtcars)
> 
> gtsummary(list(fit_1, fit_2))
Error: No glance method for objects of class lm_robust

graemeblair:
see https://github.com/vincentarelbundock/gtsummary/issues/8

lukesonnet:
Will implement ASAP (hopefully today!)

nfultz:
Extend via `generics`, same as broom - https://github.com/r-lib/generics/blob/master/R/glance.R

lukesonnet:
Yep, thanks!

lukesonnet:
Added glance for lm_robust and a Rbuildignored test for gt and gtsummary. Should I add iv_robust, and just present the second stage?

lukesonnet:
Ok we don't return those but I'm happy to inplement them. That said we
still need to decide what goes into the main body. That's what I was asking
about.

Luke Sonnet
PhD Candidate, Political Science
UCLA

graemeblair:
Sorry misunderstood, what's the main body? I thought it was one row with a set of model fit statistics.

lukesonnet:
Ah yes. That's glance. I was thinking about gtsummary and the estimates and coefficients.

lukesonnet:
Next step is to:

- mimic glance.ivreg and include standard IV metrics
- ensure that gtsummary takes the 2sls estimates in the main body
- I'll also try to do difference in means where it has basic metrics (including design type) in the glance

I'll try to do that today and then I'll open the PR.

vincentarelbundock:
Funny. I came here to see if a `glance` had been requested before only to find that my specific issue was being discussed and worked on. Thanks, I appreciate it!

Let me know if there's anything I can do to help.

lukesonnet:
Done in #285

PR #285, opened 2019-02-04
Add glance methods and IV diagnostics

lukesonnet:
In response to #282, and for general compatibility, we are adding glance methods for our four classes of estimators (lm_robust, iv_robust, difference_in_means, horvitz_thompson).

This entails adding some simple design details (N, N_blocks, design_type) for some of the non-regression estimators, and adding some diagnostics for the regression estimators, most notably 2SLS IV (Wu-Hausman, first stage F test, Sargan).

This has all been implemented, the diagnostics have been tested against ivreg and stata when possible, and sanity checks have been placed on the `glance` methods.

Notably, we don't return IV diagnostics with `fixed_effects`, warning instead, and we silently return NA for the overid tests with weights, and this is documented in `?iv_robust`.

Also see issue #287 for documenting this with gt and gtsummary.


R/S3_glance.R
```diff
@@ -0,0 +1,51 @@
...
+#' @export
+generics::glance
+
+glance_lm_robust <- function(x, ...) {
+  ret <- cbind(
+    data.frame(
+      r.squared = x[["r.squared"]],
+      adj.r.squared = x[["adj.r.squared"]]
+    ),
+    if (exists("fstatistic", x)) {
```
nfultz: Instead of ternary, maybe just put an if block at the bottom.

R/S3_print.R
```diff
@@ -56,7 +56,7 @@ print_summary_lm_like <- function(x,
     fstat
   )
 
-  if (!is.null(x$proj_fstatistic)) {
+  if (exists("proj_fstatistic", x)) {
```
nfultz: maybe is.numeric here instead of exists, to be somewhat stricter and match fstat <- line above?

R/S3_summary.R
```diff
@@ -57,23 +57,31 @@ summary.iv_robust <- function(object, ...) {
...
+      out_values,
       "proj_r.squared",
       "proj_adj.r.squared",
-      "proj_fstatistic",
-      "res_var",
-      "weighted",
-      "se_type",
-      "fes"
-    )]
+      "proj_fstatistic"
```
nfultz: was the issue that you were getting nulls in return list when no fe's in the model?

tests/testthat/test-gtsummary.R
```diff
@@ -0,0 +1,20 @@
...
+  library(gt)
+  library(gtsummary)
+
+  model1 <- lm_robust(mpg ~ am, mtcars)
+  model2 <- lm_robust(mpg ~ am, mtcars, clusters = cyl)
+  model3 <- lm_lin(mpg ~ am, ~ cyl, mtcars)
+
+  gto <- gtsummary(list(model1, model2, model3))
+
+  expect_equal(ncol(gto), 3L)
```
nfultz: Maybe paste in the actual column names here, eg to make sure ordering doesn't accidentally change when code is next touched.

tests/testthat/test-s3-methods.R
```diff
@@ -49,6 +48,18 @@ test_that("tidy, summary, and print work", {
     1
   )
 
+  capture_output(glance(lmo))
```
nfultz: expect_output may be better here?

tests/testthat/test-s3-methods.R
```diff
@@ -49,6 +48,18 @@ test_that("tidy, summary, and print work", {
...
+
+  expect_is(glance(lmo), "data.frame")
```
nfultz: instead of constantly re-calling glance(lmo) every line, create the object once and test that:

```
o <- glance(lmo)

expect_is(o, "data.frame")

```

lukesonnet:
Thanks for the comments Neal. Will implement tomorrow.

Note that in Slack we have decided not to return the first stage f-stats in `glance.iv_robust` if there are multiple endogenous regressors. See the problem posed here: https://github.com/tidymodels/broom/issues/599

I will still compute the first-stage f-stats but they will only be visible in the returned object and when you print `summary.iv_robust`, not in `glance.iv_robust`.

R/helper_lm_robust_fit.R
```diff
@@ -535,8 +530,24 @@ get_fstat <- function(tss_r2s,
   return(f)
 }
 
-compute_fstat <- function(coefs, vcov_fit, nomdf) {
-  crossprod(coefs, chol2inv(chol(vcov_fit)) %*% coefs) / nomdf
+compute_fstat <- function(coef_matrix, coef_indices, vcov_fit, rank, nomdf) {
+  fstat <- tryCatch({
+    vapply(seq_len(ncol(coef_matrix)),
```
nfultz: Not everything needs to be an apply - I would keep the original function around, then use it in a for-loop.

Also you can simplify your sandwich using backsolve.

R/helper_lm_robust_fit.R
```diff
@@ -535,8 +530,24 @@ get_fstat <- function(tss_r2s,
...
+             crossprod(
+               coef_matrix[coef_indices, x],
+               chol2inv(chol(vcov_fit[vcov_indices, vcov_indices])) %*%
+                 coef_matrix[coef_indices, x]
+             ) / nomdf
+           },
+           numeric(1))
+  },
+  error = function(e) {
+    rep(NA_real_, length(ncol(coef_matrix)))
```
nfultz: If one fails, the whole thing is NA?

lukesonnet: Hmmm, good point. Originally was written with only one outcome. Will fix and test.

R/helper_lm_robust_fit.R
lukesonnet: So I could keep the original function, but this is now used in a couple places so it would just be two functions, where one is only called by the other.

In other news, why for over vapply?

R/S3_glance.R
lukesonnet: The reason is for the expected order, I imagine (I was mostly mimicking broom). Would you prefer an if block at the bottom with a column re-ordering?

R/helper_lm_robust_fit.R
nfultz: I think a for-loop would make it clearer what you're doing - 

actually, is there any reason not to factor the block matrix directly, and vectorize the entire calculation, instead of piece by piece? Are there some zeros on the diagonal or something else that would make it fail?

R/S3_glance.R
nfultz: +1

tests/testthat/test-iv-robust.R
```diff
@@ -444,26 +464,72 @@ test_that("IV diagnostics", {
...
   }
 
-  for (f in formulae) {
+  for (f_n in names(formulae)) {
+    f <- formulae[[f_n]]
     ivro <- iv_robust(f, data = mtcars, se_type = "classical", diagnostics = TRUE)
-    aer_ivro <- summary(ivreg(f, data = mtcars), diagnostics = TRUE)
+    aer_ivro <- summary(AER::ivreg(f, data = mtcars), diagnostics = TRUE)
+
+    # Sargan stat seems to be wrong for AER for this model (-ve critical value)
```
nfultz: !?

lukesonnet: I don't know.

R/S3_summary.R
lukesonnet: Yes

lukesonnet:
This PR is ready to be merged pending code review and tests passing.

R/S3_glance.R
```diff
@@ -0,0 +1,195 @@
...
+#' @family estimatr glancers
+#' @md
+glance.difference_in_means <- function(x, ...) {
+  ret <- cbind(
+    data.frame(
+      design = x[["design"]],
+      df = x[["df"]],
+      N = as.integer(x[["N"]])
+    ),
+    if (exists("N_blocks", x)) {
```
nfultz: For these, I'd recommend wither using the coalesce operator `%||%` or a helper:

    value <- function(x. what) if(exists(what, x)) x[[what]] else NA_real_

and then instead of a cbind with if/elses, you have one data frame with

    data.frame(..., 
      N_blocks = x[["N_blocks"]] %||% NA_real_,
      ...
    )

R/S3_print.R
```diff
@@ -89,6 +101,7 @@ print.summary.lm_robust <- function(x,
 #' @export
 print.summary.iv_robust <- function(x,
                                     digits = max(3L, getOption("digits") - 3L),
+                                    signif.stars = getOption("show.signif.stars"),
```
nfultz: thread through signif.stars to print_summary_lm_like ?

R/S3_glance.R
```diff
@@ -0,0 +1,195 @@
...
+#' @export
+#' @family estimatr glancers
+#' @md
+glance.horvitz_thompson <- function(x, ...) {
+  ret <- data.frame(
+    N = as.integer(x[["N"]]),
+    se_type = x[["se_type"]],
+    condition2 = x[["condition2"]],
+    condition1 = x[["condition1"]]
+  )
```
nfultz: stringsAsFactors=FALSE ?

R/S3_print.R
```diff
@@ -76,6 +74,20 @@ print_summary_lm_like <- function(x,
   }
   cat("\n")
 
+  if (is.numeric(x[["diagnostic_endogeneity_test"]])) {
+    cat("\nDiagnostic tests:\n")
```
nfultz: Should it be "Diagnostic tests:" or just "Diagnostics:"

R/estimatr_difference_in_means.R
```diff
@@ -431,7 +443,9 @@ difference_in_means_internal <- function(condition1 = NULL,
 
   # Check that treatment status is uniform within cluster, checked here
   # so that the treatment vector t doesn't have to be built anywhere else
+  clustered <- FALSE
```
nfultz: clustered <- is.numeric(data$cluster)
    if(clustered) {
      ...
    }

R/S3_print.R
lukesonnet: Diagnostics is probably better, will change.

R/S3_glance.R
lukesonnet: Thanks, will change.

R/S3_glance.R
lukesonnet: Thanks, will change. I'd rather just use base R so I'll stick to the function.

R/estimatr_difference_in_means.R
lukesonnet: I won't use is.numeric because it need not be, but I can use !is.null

R/S3_print.R
lukesonnet: Yes!

R/S3_print.R
```diff
@@ -94,6 +94,7 @@ print_summary_lm_like <- function(x, digits, signif.stars = getOption("show.sign
 #' @export
 print.summary.lm_robust <- function(x,
                                     digits = max(3L, getOption("digits") - 3L),
+                                    signif.stars = getOption("show.signif.stars"),
```
nfultz: pass signif.stars through to print_summary_lm_like

.travis.yml
```diff
@@ -26,10 +26,12 @@ matrix:
     - os: osx
       r: release
       if: branch = master
+      brew_packages: libgit2
```
nfultz: Is travis having problems?

lukesonnet: Yes, only on mac. says it needs that package for git2r

lukesonnet: It's happening on all estimatr branches

nfultz: :( 

I would cherry pick this edit (git cherrypick --help), push that (and only that) on to master, then in the other PRs do the update from master button

lukesonnet: Got it.

lukesonnet:
This is ready to be merged in. The only failure is on the branch travis, where the changes to the tests that are now in master have not been updated. You'll notice both PR tests have passed.

Please approve and we will merge.

PR #288, opened 2019-02-26
Fix travis issues

lukesonnet:
- libgit2 needed for mac osx builds
- in R-devel, several test errors

For the latter, the problem comes from seeds generating different random data in R-devel. In two cases, this meant that the test configuration failed for non-serious reasons:

- for the predict test error, the sample data had collinearity and I just needed to increase the size or change the seed
- for the condition_pr_matrix test error, I needed to increase the sample draws of the treatment vector to make sure the condition_pr_mat built using treatment permutations still matches the analytic condition_pr_mat

However, for the fstat error, the new random data revealed that there are still differences between fstats computed via `lm_robust` and `commarobust`. This is due to (computational error) numerical differences between `lm` and `lm_robust`. For now, I'm avoiding this edge case in the test. A longer term situation is to detect when the `chol()` decomposition of the `vcov` matrix is unstable and return NA, but I'm not sure of a safe way to do that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,10 +26,12 @@ matrix:
     - os: osx
       r: release
       if: branch = master
+      brew_packages: libgit2
 
     - os: osx
       r: 3.4
       if: branch = master
+      brew_packages: libgit2
 
 env:
   global:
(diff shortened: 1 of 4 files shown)


lukesonnet:
Hacky solution to the latter, but I can't find a more principled way to fail the fstat.

lukesonnet:
What triggers the above required check?

nfultz:
> What triggers the above required check?

Commits to the pull request, or you can go in manually and restart them.

lukesonnet:
But it's just the one that seems to be different from the travis CI builds that have already passed...

Issue #293, opened 2019-03-02
fixed_effects breaks with collinearity in the covariates


lukesonnet:
This is caused by `variable_names` still having all the columns on line 263 despite collinear columns being dropped in `data[["X"]]` in line 149.

```
dat <- data.frame(
  y = rnorm(4),
  x1 = 1:4,
  x2 = 1:4,
  z = c(0,0,1,1)
)

lm_robust(y ~ x1 + x2, fixed_effects = ~ z, data = dat)
```

Output:
```
Error in data[["Xoriginal"]][, variable_names, drop = FALSE] : 
  subscript out of bounds
```

Issue #149, opened 2018-03-07
refactor `lm_robust_fit`


lukesonnet:
It is starting to look like spaghetti.

lukesonnet:
This is already a lot better thanks to PRs #209, #218, and #205. Still leaving this issue up because it could be refactored again taking a big picture view of the functions place in the package.

PR #294, opened 2019-03-02
Fix broken fes with collinear covariates

lukesonnet:
Closes issue #293. Thanks Chad.

### DIFF
--- a/R/helper_lm_robust_fit.R
+++ b/R/helper_lm_robust_fit.R
@@ -3,7 +3,7 @@
 #' @param y numeric outcome vector
 #' @param X numeric design matrix
 #' @param yoriginal numeric outcome vector, unprojected if there are fixed effects
-#' @param Xoriginal numeric design matrix, unprojected if there are fixed effects
+#' @param Xoriginal numeric design matrix, unprojected if there are fixed effects. Any column named \code{"(Intercept)" will be dropped}
 #' @param weights numeric weights vector
 #' @param cluster numeric cluster vector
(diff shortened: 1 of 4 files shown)


R/helper_lm_robust_fit.R
```diff
@@ -260,7 +260,11 @@ lm_robust_fit <- function(y,
       return_list[["fixed_effects"]] <- setNames(
         tapply(
           return_list[["fitted.values"]] -
-            data[["Xoriginal"]][, variable_names, drop = FALSE] %*% fit$beta_hat,
+            data[["Xoriginal"]][
+              ,
+              colnames(data[["X"]]),
```
nfultz: all_variable_names ? The intent would be clearer here if you updated all_variable_names at the same place where Xoriginal loses those columns. 

Let's also think about splitting up this 300 line function into something more manageable someday maybe. This is a good candidate for a helper.

lukesonnet:
https://github.com/DeclareDesign/estimatr/issues/149 :(

lukesonnet:
I think I've created a cleaner solution.

- Now we just carry around one set of variable_names, which is all of the variable names.
- Later on, we clobber the data to remove collinear columns from both X, Xoriginal, and other versions of X
- The problem was the Xoriginal still had the intercept, so I just drop that up front if Xoriginal exists, as we never need that intercept column

Agreed re: general refactor, but I think I'll ticket that as my next big item with the DoFs, and will try that near the end ofo the month.

Issue #296, opened 2019-03-05
bug in tidy.lh results in repeating rows for multiple linear hypotheses


lukesonnet:
`tidy(lh_robust(mpg ~ cyl + am, data = mtcars, linear_hypothesis = c("cyl=1", "cyl=am"))$lh)`
Output:
```
    term  estimate std.error  statistic      p.value  conf.low conf.high df outcome
1  cyl=1 -3.500958 0.3114269 -11.241666 4.349494e-12 -4.137897 -2.864018 29     mpg
2 cyl=am -5.067992 1.0643572  -4.761552 4.919824e-05 -7.244847 -2.891137 29     mpg
3  cyl=1 -3.500958 0.3114269 -11.241666 4.349494e-12 -4.137897 -2.864018 29     mpg
4 cyl=am -5.067992 1.0643572  -4.761552 4.919824e-05 -7.244847 -2.891137 29     mpg
```

lukesonnet:
This is caused by the way we repeat `outcome` to accommodate multiple outcomes in multivariate regressions. I believe the solution is to just break the `lh` data.frame into a list and clobber the outcome to be `unique(outcome)` when we tidy or print an `lh` object.

Otherwise, we could have `lh` be a list instead of a data.frame, and turn it into a data.frame with `tidy.lh`, but it's probably too late to change that.

Issue #295, opened 2019-03-05
bug in summary.lh_robust() for linear hypotheses


lukesonnet:
```
> library(estimatr)
> summary(lh_robust(mpg ~ am, data = mtcars, linear_hypothesis = "am=2"))
Loading required namespace: car
$lm_robust 
 
Call:
lh_robust(mpg ~ am, data = mtcars, linear_hypothesis = "am=2")

Standard error type:  HC2 

Coefficients:
            Estimate Std. Error t value  Pr(>|t|) CI Lower CI Upper DF
(Intercept)   17.147     0.8796  19.495 1.378e-18   15.351    18.94 30
am             7.245     1.9232   3.767 7.211e-04    3.317    11.17 30

Multiple R-squared:  0.3598 ,	Adjusted R-squared:  0.3385 
F-statistic: 14.19 on 1 and 30 DF,  p-value: 0.0007211


$lh 
 
NULL
```

Note the NULL, which is due to `attr(x, "linear_hypothesis")` not existing.

lukesonnet:
Our two options here are either to use `stats:::print.anova()` which the above may be attempting to do, or to just print the lh data.frame out, which I think is preferable.

Option 1: stats:::print.anova():
```
$lm_robust

Call:
lh_robust(mpg ~ am, data = mtcars, linear_hypothesis = "am=2")

Standard error type:  HC2 

Coefficients:
            Estimate Std. Error t value  Pr(>|t|) CI Lower CI Upper DF
(Intercept)   17.147     0.8796  19.495 1.378e-18   15.351    18.94 30
am             7.245     1.9232   3.767 7.211e-04    3.317    11.17 30

Multiple R-squared:  0.3598 ,	Adjusted R-squared:  0.3385 
F-statistic: 14.19 on 1 and 30 DF,  p-value: 0.0007211

$lh
Linear hypothesis test

Hypothesis:
am = 2

Model 1: restricted model
Model 2: mpg ~ am

  Res.Df Df  Chisq Pr(>Chisq)   
1     31                        
2     30  1 8.8364   0.002953 **
---
Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
```

Option 2:
```
$lm_robust

Call:
lh_robust(mpg ~ am, data = mtcars, linear_hypothesis = "am=2")

Standard error type:  HC2 

Coefficients:
[...]

PR #298, opened 2019-03-05
Add glance.lh_robust; add tests and bugfixes for lh_robust s3 methods

lukesonnet:
Issues #295, #296


acoppock:
These changes look sensible to me. Any sense why it's not passing?

PR #313, opened 2019-07-22
Nfultz/fix solaris3

### LINKED ISSUE #268: Fix segfault issue in demeanMat
This error comes up during a couple calls to `demeanMat` on Solaris (https://www.r-project.org/nosvn/R.check/r-patched-solaris-x86/estimatr-00check.html).

Seems to mostly be triggered by `iv_robust()`.

nfultz:
This fixes the segfault issue on solaris.

closes #268


lukesonnet:
How did you manage to pinpoint the problem? I couldn't even isolate it in debugging.

lukesonnet:
@nfultz, please update ctb to aut and I'll approve (it'll also trigger what will hopefully be a fixed travis PR run!

lukesonnet:
Have you done any speed tests on this? I wonder whether `match` is slow, and whether `fmatch` makes a difference? If not I can take a look at this later.

nfultz:
> @nfultz, please update ctb to aut and I'll approve (it'll also trigger what will hopefully be a fixed travis PR run!

https://github.com/nfultz/estimatr/pull/1 <- like this?

> Have you done any speed tests on this? I wonder whether match is slow, and whether fmatch makes a difference? If not I can take a look at this later.

The net change of the full patch should be a speed increase in practice if not in the big O / worst case, because the indices can be reused between X, Y, and Z instead of being recalculated for each. 

There shouldn't be any need to pull in fmatch, because we are manually doing the book keeping to reuse those indexes, which is the trick fmatch uses for speed afaict.

PR #339, opened 2020-02-04
Starprep list

jaspercooper:
Thanks to @notanastronomer for the heads up on this! I've edited her MWE below a bit. 

People often like to carry around lists of lm fits, e.g. using lapply over a list of formulas. Currently, these are not handled by `starprep()`.

```
library(estimatr)
library(stargazer)
a <- lm(mpg ~ cyl, mtcars)
b <- lm(mpg ~ cyl + disp, mtcars)
c <- lm(mpg ~ cyl + disp + hp, mtcars)

abc <- list(a, b, c)
ab <- list(a, b)

# Works
starprep(a,b,c)
# Breaks
starprep(abc)
# Works
starprep(a,b)
# Breaks
starprep(ab)
```

This behavior worries me a bit, as I can see people doing something like this:

```
stargazer(a, b, se = starprep(ab), type = "text")
```

Expecting to get this:

```
stargazer(a, b, se = starprep(a,b), type = "text")
```

I've written an admittedly very quick and dirty fix in this PR, in case it's helpful to build off.

### DIFF
--- a/R/helper_starprep.R
+++ b/R/helper_starprep.R
@@ -252,14 +252,19 @@ starprep <- function(...,
 
   fits <- list(...)
 
-  gave_list <- length(fits) == 1 & is.list(fits[[1]])
+  first_object <- fits[[1]]
 
-  if(gave_list) fits <- fits[[1]]
+  gave_list <- length(fits) == 1 & is.list(first_object)
+
+  if(gave_list){
+    is_list_of_lm <- sapply(first_object, inherits, what = c("lm","lm_robust"))
+    if(all(is_list_of_lm)) fits <- first_object
+  }
 
   fitlist <- lapply(
     fits,
     function(x) {
-      if (class(x)[1] == "lm")
+      if (inherits(x, "lm"))
         commarobust(x, se_type = se_type, clusters = clusters, alpha = alpha)
       else
         x
--- a/R/helper_starprep.R
+++ b/R/helper_starprep.R
@@ -249,8 +249,15 @@ starprep <- function(...,
                      se_type = NULL,
                      clusters = NULL,
                      alpha = 0.05) {
+
+  fits <- list(...)
+
+  gave_list <- length(fits) == 1 & is.list(fits[[1]])
+
+  if(gave_list) fits <- fits[[1]]
+
   fitlist <- lapply(
-    list(...),
+    fits,
     function(x) {
       if (class(x)[1] == "lm")
         commarobust(x, se_type = se_type, clusters = clusters, alpha = alpha)
--- a/tests/testthat/test-starprep.R
+++ b/tests/testthat/test-starprep.R
@@ -248,3 +248,20 @@ test_that("Only works with lm, not mlm or glm", {
   )
 })
 
+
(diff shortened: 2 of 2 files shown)


R/helper_starprep.R
```diff
@@ -249,8 +249,15 @@ starprep <- function(...,
                      se_type = NULL,
                      clusters = NULL,
                      alpha = 0.05) {
+
+  fits <- list(...)
+
+  gave_list <- length(fits) == 1 & is.list(fits[[1]])
```
nfultz: I think this code is probably correct, but unclear - I would double check that `is.list` is false for lm objects - you don't want to break the case where only one model is provided.

    is.list(lm(extra~group, sleep))

Also while you are in here, can you change the `class(x)[1] == "lm"` below to use `inherits`.

jaspercooper: oof, good point. will update now

jaspercooper:
@nfultz addressed the two comments on last commit, please lmk if this seems clearer and less error-prone

R/helper_starprep.R
```diff
@@ -249,10 +249,22 @@ starprep <- function(...,
                      se_type = NULL,
                      clusters = NULL,
                      alpha = 0.05) {
+
+  fits <- list(...)
+
+  first_object <- fits[[1]]
+
+  gave_list <- length(fits) == 1 & is.list(first_object)
```
nfultz: I think it might actually better to check if the first dots is an lm or not, instead of trying to write logic to detect a list - see also https://github.com/DeclareDesign/DeclareDesign/blob/3965a5c56631ecec2fc6ad05be28a7a196213b31/R/diagnose_design.R#L83 in DeclareDesign where there's ambiguity / similar logic bw data.frames and lists of data frames.

R/helper_starprep.R
```diff
@@ -249,10 +249,22 @@ starprep <- function(...,
...
+    is_list_of_lm <- sapply(first_object, inherits, what = c("lm","lm_robust"))
+    if(all(is_list_of_lm)) fits <- first_object
+  }
+
   fitlist <- lapply(
-    list(...),
+    fits,
     function(x) {
-      if (class(x)[1] == "lm")
+      if (inherits(x, "lm"))
```
nfultz: +1 - also what's the point of the `else x` - when would that ever be desired.

lukesonnet: when x is "lm_robust", no?

R/helper_starprep.R
lukesonnet: I think it's better to cast as fits first so that we can sanitize input. I'll PR some changes in a second.

lukesonnet: The main problem is that lms and lm_robusts are also lists, right? I'll PR a change.

graemeblair:
Requesting review @nfultz @lukesonnet. I merged in Luke's PR to this one.

R/helper_starprep.R
nfultz: OK LGTM.

nfultz:
LGTM.

PR #340, opened 2020-02-05
change list checking logic and add some checks

lukesonnet:
This just updates @jaspercooper's work on allowing lists to starprep, adding a couple of error checks and some safer type-checking,

### DIFF
--- a/R/helper_starprep.R
+++ b/R/helper_starprep.R
@@ -250,15 +250,19 @@ starprep <- function(...,
                      clusters = NULL,
                      alpha = 0.05) {
 
-  fits <- list(...)
+  if (inherits(..1, "list")) {
+    if (...length() > 1) {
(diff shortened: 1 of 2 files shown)


R/helper_starprep.R
```diff
@@ -250,15 +250,19 @@ starprep <- function(...,
                      clusters = NULL,
                      alpha = 0.05) {
 
-  fits <- list(...)
-
-  first_object <- fits[[1]]
+  if (inherits(..1, "list")) {
```
nfultz: This works because although a lm is.list it does not inherit, which is a little nuanced

```
> is.list(lm(extra~group, sleep))
[1] TRUE
> inherits(lm(extra~group, sleep), "list")
[1] FALSE
```

Id recommend putting the lm case first

    if(inherits(..1, 'lm') { } else if (inherits(..1, "list") && ...length() == 1) {} else stop()

PR #341, opened 2020-02-05
Add n clusters to extract.lm_robust

lukesonnet:
While `extract.lm_robust` is gross, it looks like the other extract functions. So until such a time it gets completely re-written, I'm just going to add features as in this PR, where I add the ability to include the number of clusters for clustered designs.

### DIFF
--- a/tests/testthat/test-texreg.R
+++ b/tests/testthat/test-texreg.R
@@ -6,7 +6,7 @@ test_that("texreg extension works", {
 
   model2 <- lm_robust(extra~group, sleep, clusters = ID)
 
-  capture.output(treg <- extract(model2, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE))
+  capture.output(treg <- extract(model2, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE))
   expect_is(
     treg,
     "texreg"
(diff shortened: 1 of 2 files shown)


tests/testthat/test-texreg.R
```diff
@@ -6,7 +6,7 @@ test_that("texreg extension works", {
 
   model2 <- lm_robust(extra~group, sleep, clusters = ID)
 
-  capture.output(treg <- extract(model2, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE))
+  capture.output(treg <- extract(model2, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE))
```
nfultz: Maybe use named arguments here?

    extract(model2, include.fstatistic = TRUE)

tests/testthat/test-texreg.R
```diff
@@ -18,4 +18,7 @@ test_that("texreg extension works", {
   # Remove to get SEs
   expect_true(grepl("\\(0.57\\)", texreg::texreg(model2, include.ci = FALSE)))
 
+  # Has Nclust
+  expect_true(grepl("N Clusters.*10", texreg::texreg(model2)))
```
nfultz: You can narrow the test by using `\s` to match whitespace

PR #346, opened 2020-04-15
add modelsummary to tests, regression table vignette, and DESCRIPTION

vincentarelbundock:
This PR supersedes this WIP: https://github.com/DeclareDesign/estimatr/pull/326

`modelsummary` and `gt` are now on CRAN, so we can add a test and a note in the vignette.

The test raises a warning because some of the `estimatr` objects do not have an associated `nobs` method. I could try to put one together in a separate PR if you think that's useful.

PR #347, opened 2020-04-15
new nobs methods and standardize glance output

vincentarelbundock:
New `nobs` methods for `horvitz_thompson` and `difference_in_means` objects

Normalize `glance` output to conform to usual `broom` nomenclature:

* N -> nobs
* N_clusters -> nclusters
* N_blocks -> nblocks

`nobs` is standard in `broom`: https://github.com/tidymodels/broom/blob/master/R/stats-lm-tidiers.R#L177

`glance` columns should be lowercase. In practice, most do not have underscores, but there is discussion here:

https://github.com/tidymodels/broom/issues/452


nfultz:
Does this have a downstream impact on DeclareDesign? If only glance has column names changed it should be ok.

vincentarelbundock:
I don't know the ecosystem well enough to be 100%, but I cloned the `DeclareDesign`, `fabricatr`, and `randomizr` repos, and `grep` can't find any call to `glance` in those packages.

vincentarelbundock:
Both of my PRs (#346 and #347) used to fail on Travis. Sorry it took me so long to get back to them!

Turns out there was nothing wrong with the code. I just merged upstream master and all the checks are now green.

[this post is copied in both PRs]

lukesonnet:
I think this is the right call, but I don't love that this isn't what is in an `lm_robust` objects output to begin with.

My pref is to just change the names in the `lm_robust` object, unless that breaks DeclareDesign.

vincentarelbundock:
I'm happy to give the more fundamental change a shot in a new PR. 

> "unless that breaks DeclareDesign"

Would making the changes and running the DeclareDesign test suite be sufficient to ensure that this is satisfied?

nfultz:
> I'm happy to give the more fundamental change a shot in a new PR.
> 
> > "unless that breaks DeclareDesign"
> 
> Would making the changes and running the DeclareDesign test suite be sufficient to ensure that this is satisfied?

Yes, that should be sufficient.  DeclareDesign master should be routing through `tidy.lm_robust` anyway, so I wouldn't expect changes to `glance` to impact it. 

@lukesonnet  Changing the name on the object for N should be fine. `lm` doesn't even store N, see also `stats:::nobs.lm`.  (I'm not sure if lm-compatibility is something people still want anymore but it was one of the original goals). But glance should probably still go through nobs() to handle weights etc?  

Also to the extent that difference_in_means  and horvitz_thompson objects are special cases of lm_robust, we could just append a class eg `class(x) < c('difference_in_means', 'lm_robust')` and let the object dispatch to nobs.lm_robust instead of implementing duplicate methods - this might simplify some other S3 stuff as well.

lukesonnet:
@nfultz, I'm sympathetic to a global class to simplify some of the code, but a little worried about unintended consequences. I think I don't want to go through that process right now.

PR #354, opened 2020-07-04
new tidy arguments: conf.int & conf.level

vincentarelbundock:
The `tidy` methods packaged with `broom` and `broom.mixed` include two arguments that are currently unsupported by `estimatr`: `conf.int` and `conf.level`. 

I ran into this problem when using `modelsummary`, a package which extracts information from models using `tidy`. This code produces a regression table and a coefficients plot, but the confidence intervals are incorrect, because `estimatr::tidy` ignores arguments that standard `tidy` methods accept:

```r
library(modelsummary)
library(estimatr)
mod <- lm_robust(mpg ~ hp + factor(cyl), mtcars)

modelsummary(mod, statistic = 'conf.int', conf_level = .99)

modelplot(mod, statistic = 'conf.int', conf_level = .99)
```

This is especially problematic because `estimatr::tidy` accepts the ellipsis, which means we can pass `conf.level=.99` and it will still return alpha=.05 without a warning.

This PR adds the missing arguments, and re-calculates CIs if the alpha level requested by `tidy` does not match the alpha level requested in the initial call to `lm_robust` et al.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,7 +23,7 @@ Imports:
     rlang (>= 0.2.0)
 LinkingTo: Rcpp, RcppEigen
 Encoding: UTF-8
-RoxygenNote: 7.1.0
+RoxygenNote: 7.1.1
 LazyData: true
 Suggests:
     fabricatr (>= 0.10.0),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# estimatr 0.23.0
+
+* tidy: rename nobs, nclusters, nblocks
+* tidy: new arguments conf.int, conf.level
+
 # estimatr 0.22.0 
 
 * Bug fixes
--- a/R/S3_tidy.R
+++ b/R/S3_tidy.R
@@ -29,14 +29,15 @@ tidy_data_frame <- function(x,
   rownames(return_frame) <- NULL
 
   # re-calculate CIs if explicitly requested and alpha level has changed
-  flag <- conf.int && utils::hasName(x, 'alpha') && (1 - x$alpha != conf.level)
+  # x$alpha[1] because lm_robust duplicates alpha with multiple linear_hypotheses
+  flag <- conf.int || (utils::hasName(x, 'alpha') && (1 - x$alpha[1] != conf.level))
   if (flag) {
       ci <- stats::confint(x, level = conf.level, ...)
       if (all(row.names(ci) == return_frame$term))  {
           return_frame$conf.low <- ci[, 1]
           return_frame$conf.high <- ci[, 2]
       }
-  }               
+  }
 
 
   return(return_frame)
--- a/R/S3_tidy.R
+++ b/R/S3_tidy.R
@@ -31,13 +31,13 @@ tidy_data_frame <- function(x,
   # re-calculate CIs if explicitly requested and alpha level has changed
   flag <- conf.int && utils::hasName(x, 'alpha') && (1 - x$alpha != conf.level)
   if (flag) {
-      ci <- tryCatch(stats::confint(x, level = conf.level), error = function(e) NULL)
-      flag <- inherits(ci, 'matrix') && all(row.names(ci) == return_frame$term)
-      if (flag) {
+      ci <- stats::confint(x, level = conf.level, ...)
+      if (all(row.names(ci) == return_frame$term))  {
           return_frame$conf.low <- ci[, 1]
           return_frame$conf.high <- ci[, 2]
       }
   }               
+
 
   return(return_frame)
 }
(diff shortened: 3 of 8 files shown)


vincentarelbundock:
This does not appear to create failures when I run the `DeclareDesign` tests.

vincentarelbundock:
Note to self: the error does not appear when I run `devtools::test` on my local machine. Now trying to figure out what this is:

```
  == testthat results  ===========================================================
  [ OK: 1330 | SKIPPED: 0 | WARNINGS: 0 | FAILED: 1 ]
  1. Error: tidy, glance, summary, and print work (@test-s3-methods.R#192) 
  
  Error: testthat unit tests failed
  Execution halted
** running tests for arch 'x64' ...
  Running 'sleep.R'
  Comparing 'sleep.Rout' to 'sleep.Rout.save' ... OK
  Running 'testthat.R' [27s]
 ERROR
Running the tests in 'tests/testthat.R' failed.
Last 13 lines of output:
   ----------- END OF FAILURE REPORT -------------- 
  -- 1. Error: tidy, glance, summary, and print work (@test-s3-methods.R#192)  ---
  'length(x) = 2 > 1' in coercion to 'logical(1)'
  Backtrace:
   1. generics::tidy(lho)
   2. estimatr:::tidy.lh_robust(lho)
   6. estimatr:::tidy.lh(x$lh)
   7. estimatr:::tidy_data_frame(simplify_lh_outcome(x))
  
  == testthat results  ===========================================================
  [ OK: 1330 | SKIPPED: 0 | WARNINGS: 0 | FAILED: 1 ]
```

vincentarelbundock:
I force-pushed an update to fix an earlier problem. Should be OK now.

lukesonnet:
Thanks again for continuing to support this. Looks good to me, but two questions.

1) why the `tryCatch`?
2) why hardcode the test values?

Thanks.

nfultz:
Now that `...` might actually do something, should also pass it through in the lh_robust tidy functions.

vincentarelbundock:
>     1. why the `tryCatch`?

Me being overly cautious. I don't know all the `estimatr` object types, so if you're confident that none will error when calling `confint`, I can remove the `tryCatch`.

>     2. why hardcode the test values?

What's the alternative you have in mind? Compare with the output of `confint`? Not sure it makes much of a difference, but easy to change if that's your preference.

> Now that `...` might actually do something, should also pass it through in the lh_robust tidy functions.

Personally, I try to avoid `...` unless there is a compelling use-case, because I find it useful for users to have immediate feedback on badly misspelled or unsupported arguments, instead of sending them into a black hole and returning something that may not match user expectations. Admittedly, this is more aspirational than anything else (I break the rule all the time), and you may have a compelling use-case in mind. Just let me know what you prefer.

nfultz:
That's a fair point, but especially for little wrapper functions, it's nice to use ... so that you only have to update the code in one spot. Also I think we'd rather not have to push a new update each time a third party dependency adds or changes a feature (in a non-breaking way), just from an ops perspective.

vincentarelbundock:
https://github.com/DeclareDesign/estimatr/pull/354/commits/bdd377c94513bcc359222cb987cf8ebfa9feec57 does:

1. Remove `tryCatch` and trust `confint`
2. Hard-coded tests now match with output of `confint`
3. ... are passed through

One case I should flag is when users call:

```r
tidy(mod, conf.level = .90)
```

In that case, `tidy` will still return the 95% interval, because the default argument is `conf.int=FALSE` (as in `broom` and elsewhere). This is because there already is a CI embedded in the object at fit time.

Is this a problem? Do you want a warning? Should the `conf.int` flag be switched to TRUE automatically when `conf.int != (1 - mod$alpha)`?

lukesonnet:
Hi Vincent, thanks again. Two comments:

> Is this a problem? Do you want a warning? Should the conf.int flag be switched to TRUE automatically when conf.level != (1 - mod$alpha)?

I think the latter. Just check that and if that or the flag is true, recompute.

> Remove tryCatch and trust confint

So I looked at `stats::confint`. It only relies on two things from the `x` object, `coef()` and `vcov()`. Right now horvitz thompson and difference in means error when `vcov()` is called on them because it's not really a well defined quantity. Instead, I think what I'll do is have `vcov.horvitz_thompson()` and `vcov.difference_in_means()` return the estimated variance as a 1,1 matrix, which I think makes the most sense, and isn't totally crazy. Then your solution will work for all of the estimatr objects.

I'll get to this later today or tomorrow, hopefully.

vincentarelbundock:
Sounds great. I'm heading to the woods for a few days. Will look at this when I come back.

vincentarelbundock:
CIs are now re-computed if `conf.int=TRUE` OR if `1-x$alpha[1] != conf.level`. Note that I added `[1]` to alpha because objects produced by `lm_robust` with 2 linear hypotheses store alpha as a vector length 2, where alpha levels seem to be duplicated. This model breaks if we don't use `x$alpha[1]`:

https://github.com/DeclareDesign/estimatr/blob/master/tests/testthat/test-s3-methods.R#L186

lukesonnet:
Please see PR on your fork for the vcov update.

vincentarelbundock:
thanks a lot. merged it.

PR #356, opened 2020-07-20
Add formula to iv_robust and add update.iv_robust

### LINKED ISSUE #353: Updating formulas does not work with iv_robust
```r
m1 <- iv_robust(Sepal.Length ~ Sepal.Width | Petal.Length, iris)

# Upon wanting to update the formula by, say, exchanging Sepal.Width for Petal.Width
m2 <- update(m1, . ~ Petal.Width | .)
# Error in clean_model_data(data = data, datargs, estimator = "iv") : 
#  Must specify a `formula` with both regressors and instruments. For example, `formula = y ~ x1 + x2 | x1 + z2` where x1 and x2 are the regressors and z1 and z2 are the instruments.
#
# See ?iv_robust.
```

And so is the case whichever way you try to update it (explictly writing the whole formula again, writing both regressor and instrument, etc.). By the way, there's a small typo in that error message, as it should have `z1` instead of `x1` after `|`.

lukesonnet:
This closes #353 

There were two problems:
1) the `terms` cannot fully capture the iv formula, and so we need to return the `formula` as is
2) We need a bespoke version of `update.default()` that works with `Formula` objects since that's how we handle instruments in our formulas. It is a copy of `update.default()` but with one line changed to use `update.Formula` instead of `update.formula`.

Added a test as well.


nfultz:
update should work out of box automatically. ?

EDIT

Since `Formula` is a dependency and `stats::formula` is a generic, I think it would be cleaner to implement `formula.iv_robust` and return a `Formula` from there, rather than editing `update` .

Since that is kinda gibberish, in code, eg

```
formula.iv_robust <- function(x, ...){
  x$formula
}

```


EDIT 2

Oh I see, `update.default` is hardcoded to use `update.formula` instead of the generic `update` :(

This breaks the drop-in-replacement-ness ability of Formula.

I'll open a ticket on R Bugzilla - it's a trivial change, and will open a different ticket to fix this someday in 4.1 or whenever.

nfultz:
See also https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17865

lukesonnet:
Thanks Neal. I didn't even think of just getting a custom `formula.iv_robust()` but as you point out it wouldn't work in this case. Thanks for filing the bug!

nfultz:
The root cause has been fixed for update in R-devel here - https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17865

When the next release of R actually ships, we should be able to add a deprecation messages to update.iv_robust and conditionally register it with S3 only for R <= 4.0. 

Note that a related issue may remain for tools that go through add / drop instead of using update directly - stepwise selection and Type III Anova come to mind. Caveat emptor.

lukesonnet:
Thanks fultzy and graeme!

PR #369, opened 2020-12-05
0.30.0 - ripley comments on suggests

acoppock:


graemeblair:
This will be posted when CRAN reopens

Issue #380, opened 2021-04-09
FR: Support quoted variable names or formulas for `clusters` and `weights`


rossellhayes:
Because functions like `lm_robust()` only support unquoted variable names for `clusters` and `weights`, it's difficult to pass arguments into them from other functions. For example:
``` r
purrr::map(
  list(am, cyl),
  ~ estimatr::lm_robust(mpg ~ hp, data = mtcars, clusters = .)
)
#> Error in purrr::map(list(am, cyl), ~estimatr::lm_robust(mpg ~ hp, data = mtcars, : object 'am' not found
```

<sup>Created on 2021-04-09 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

We get an error that 'am' cannot be found because `list()` tries to evaluate its arguments before they can reach `lm_robust()`.

If `lm_robust()` could accept its arguments in a form that `list()` did not try to evaluate, like a quoted name or a formula, this would solve the problem.

nfultz:
Because estimatr has a frankenstein of base R and rlang NSE, this is hard to get working without reading the actual code.

But this seems to work for me:

```
purrr::map2(
  list(mpg ~ hp, mpg ~ wt),
  list(quo(am), quo(cyl)),
  ~ lm_robust(.x, data = mtcars, clusters = !!.y)
)
```

Or with your new example,

```
purrr::map(
  quos(am, cyl),
  ~ estimatr::lm_robust(mpg ~ hp, data = mtcars, clusters = !!.)
)
```

`quo`/`quos` is used in the tidy-eval framework to delay the evaluation until the model frame is constructed. I would recommend updating the documentation to indicate quosures are useable, rather than changing the code to recursively evaluate formulas.

rossellhayes:
Thanks, this is a really helpful solution!

PR #381, opened 2021-04-10
Add support for right-sided formulas as arguments `clusters` and `weights`

rossellhayes:
This PR addresses #380 by adding support for right-sided formulas as arguments `clusters` and `weights` in model functions. With this change, `clusters` and `weights` can more easily be set by other functions calling `estimatr` models, solving the motivating problem in #380:

``` r
purrr::map(
  list(~ am, ~ cyl),
  ~ estimatr::lm_robust(mpg ~ hp, data = mtcars, clusters = .)
)
#> [[1]]
#>                Estimate  Std. Error   t value   Pr(>|t|)   CI Lower    CI Upper
#> (Intercept) 30.09886054 2.450272545  12.28388 0.05171030 -1.0340203 61.23174140
#> hp          -0.06822828 0.004484306 -15.21490 0.03879625 -0.1215026 -0.01495395
#>                   DF
#> (Intercept) 1.000011
#> hp          1.029391
#> 
#> [[2]]
#>                Estimate Std. Error   t value   Pr(>|t|)   CI Lower  CI Upper
#> (Intercept) 30.09886054 5.04648516  5.964322 0.05856662 -3.7089449 63.906666
#> hp          -0.06822828 0.02652462 -2.572262 0.17539476 -0.2460305  0.109574
#>                   DF
#> (Intercept) 1.394247
#> hp          1.393706
```

<sup>Created on 2021-04-09 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>


R/helper_clean_model_data.R
```diff
@@ -44,6 +44,17 @@ clean_model_data <- function(data, datargs, estimator = "") {
   for (da in to_process) {
     name <- sprintf(".__%s%%%d__", da, sample.int(.Machine$integer.max, 1))
     m_formula_env[[name]] <- eval_tidy(mfargs[[da]], data = data)
+
+    if (is_formula(m_formula_env[[name]])) {
```
nfultz, line 48: I think this affects all args, not just weights and clusters? Not sure if that is bad though. Maybe move outside loop like fixed effects?

nfultz:
See also my comment on the ticket - if the workaround is sufficient, no need to merge  - https://github.com/DeclareDesign/estimatr/issues/380#issuecomment-817151396

rossellhayes:
Agreed, using quos was able to handle my use case, so I think that should be sufficient.

R/helper_clean_model_data.R
rossellhayes: This may be moot based on the comments below, but this is a good point. I don't think this will have negative consequences, since none of the arguments that go through this loop should evaluate to a formula, but there may be a speed penalty from adding an unnecessary step to every loop.

PR #382, opened 2021-04-11
fix typo in stata-wls-hat.Rmd

elbersb:
Sorry, the second parenthesis was still in the wrong place... Now it should be correct.


lukesonnet (approved):
Oof! good catch! Thanks, and sorry for the late response.

Issue #403, opened 2023-03-13
predict and residuals have odd behavior


lmiratrix:
Check out
```

dat = tibble( siteID = rep( LETTERS[1:4], each=10 ),
              X = rnorm( 40 ),
              Y = X + rnorm( 40 ) )
mod = lm_robust( Y ~ X + siteID, data=dat )
mod
predict( mod )
residuals( mod )
```

Should these return values?

nfultz:
See also #123

acoppock:
Thanks for reporting this.

Both now return values: `predict()` with no `newdata` gives the in-sample fit, and `residuals()` gives the residuals, each matching `lm()`.

``` r
packageVersion("estimatr")
#> [1] '2.0.0'
library(estimatr)
set.seed(343)
dat <- data.frame(siteID = rep(LETTERS[1:4], each = 10), X = rnorm(40))
dat$Y <- dat$X + rnorm(40)

mod <- lm_robust(Y ~ X + siteID, data = dat)
head(predict(mod))
#>          1          2          3          4          5          6 
#> -1.2126424  0.4705953  0.1365516 -0.7303746 -0.3491856 -1.2724810
head(residuals(mod))
#> [1]  0.55436298 -0.83549723  0.75101852 -0.40273711 -0.61998766  0.06108392

ols <- lm(Y ~ X + siteID, data = dat)
all.equal(unname(predict(mod)), unname(fitted(ols)))
#> [1] TRUE
all.equal(unname(residuals(mod)), unname(residuals(ols)))
#> [1] TRUE
```

The fix is in estimatr 2.0.0, which is on `main` and has been submitted to CRAN; until CRAN accepts it, `remotes::install_github("DeclareDesign/estimatr")` installs it.

PR #416, opened 2025-01-07
Prediction with lm_lin() fixes #415

### LINKED ISSUE #415: Revisions to prediction with lm_lin()
Currently prediction does not work for `lm_lin()` with multi-valued or factorial treatments. This is because of how `predict.lm_robust()` handles generation of the lin estimator model matrix with new data.  The treatment name saved in the `lm_lin()` model object refers to the original variable name, which may have been transformed in the model matrix to multiple columns, causing some disagreement when treatment x covariate interactions are created. The original variable name doesn't exist in the revised model matrix, and/or the new data model matrix doesn't have correct dimensions when multiplied by coefficients. 

See [here](https://github.com/DeclareDesign/estimatr/blob/521a331bf4afe5a8e9e59a649545d137e37d655c/R/S3_predict.R#L91). 

For example:
```
library(estimatr)
set.seed(60637)

N <- 40
dat <- data.frame(
  x = rnorm(N, mean = 2.3),
  x2 = rpois(N, lambda = 2),
  x3 = runif(N)
)

dat$y0 <- rnorm(N) + dat$x
dat$y1 <- dat$y0 + 0.35
dat$y2 <- dat$y0 + 0.55

dat$z_multi <- sample(0:2, size = nrow(dat), replace = TRUE)
dat$z_bin <- 1*(dat$z_multi>0)
dat$y <- (dat$z_multi == 0)*dat$y0 + (dat$z_multi == 1)*dat$y1 + (dat$z_multi == 2)*dat$y2
```

```
# Multi-valued numeric treatments with lm_lin; estimation works as expected
lmlin_mult <- lm_lin(y ~ z_multi, covariates = ~ x, data = dat)
# prediction does not
predict(lmlin_mult, newdata = dat)
# Error in X[, !beta_na, drop = FALSE] :
#   (subscript) logical subscript too long

# Binary factorial treatment with lm_lin; estimation works,
lmlin_bin_f <- lm_lin(y ~ as.factor(z_bin), covariates = ~ x + x2 + x3, data = dat)
# prediction breaks
predict(lmlin_bin_f, newdata = dat)
# Error in X[, treat_name] : subscript out of bounds
```

More detail in gist [here](https://gist.github.com/mollyow/cbdcdc44317a09f1b2b268ad737380a4)

A revision to handle setting up treatment columns in the new data could be implemented in `get_X()`.

mollyow:
This PR modifies `get_X()` in `predict.lm_robust()` to appropriately create model matrices for new data predicted from `lm_lin()` models with multi-valued and factorial treatments, and fixes #415. 

Changes in this PR:
- allows for prediction with `lm_lin()` when treatment is a factor and/or multi-valued **(primary goal)**
    - adds saved `treatment_levels` to the returned `lm_lin` model object
    - stops prediction for `lm_lin` if the treatment values in new data are not a subset of `treatment_levels`. <sup>1</sup>
- standardizes model fit for `lm_lin()` models with no intercept <sup>2</sup>
- adds tests to ensure identical predictions from `lm_lin()` models where treatment is either numeric or factorial, and fit with/without an intercept (there may be extremely small floating point differences)
- adds relevant examples to `predict.lm_robust` and `lm_lin` documentation. 

-----
Notes:
1: This change has consequences for use with `margins::margins()`, which now must be instructed that treatment is a factor. 
 - Why? `margins()` perturbs values of the variable to get marginal effects. The perturbed values will not be a subset of the original treatment levels, which now throws an error. 
 - If we only ever had binary treatments, this would not be an issue. However, `lm_lin()` allows users to input multi-valued treatments as a numeric variable, but recognizes each distinct numeric value as a different treatment level. Meanwhile, `margins()` would still treat these variables as continuous, resulting in different errors or weird behavior. The `margins()` package already has intended behavior for factor variables following Stata implementation ([`margins` #6](https://github.com/leeper/margins/issues/6)), users just need to explicitly instruct margins that treatment is a factor to get correct behavior. 

2: 
There is a small change to modify behavior of `lm_lin()` for binary 0/1 treatments with no intercept. I think it's an open question what correct behavior should be here, as in Winston's original paper all models have intercepts. Previously, with binary treatment, if there's no intercept you would get a model with a treatment indicator, de-meaned covariates, and treatment interacted with covariates:

```
set.seed(60637)

N <- 40
dat <- data.frame(
  x = rnorm(N, mean = 2.3),
  x2 = rpois(N, lambda = 2),
  x3 = runif(N)
)

dat$y0 <- rnorm(N) + dat$x
dat$y1 <- dat$y0 + 0.35

dat$z_bin <- sample(0:1, size = nrow(dat), replace = TRUE)
dat$y <- (dat$z_bin == 0)*dat$y0 + (dat$z_bin == 1)*dat$y1

# Binary 01 treatments with lm_lin and no intercept
lmlin_bin <- lm_lin(y ~ z_bin-1, covariates = ~ x, data = dat)
```
Produces:
```
           Estimate Std. Error   t value     Pr(>|t|)   CI Lower CI Upper DF
z_bin     2.5165919  0.2647700 9.5048223 1.797926e-11  1.9801169 3.053067 37
x_c       0.6103258  0.6999342 0.8719759 3.888449e-01 -0.8078756 2.028527 37
z_bin:x_c 0.5577459  0.7818820 0.7133377 4.801122e-01 -1.0264974 2.141989 37
```
This is difficult to interpret in terms of a treatment effect. 

This PR changes behavior for binary 0/1 treatment to be the same as what you would see when treatment is multi-valued or otherwise treated as a factor. It also allows you to back out the Lin estimate of the ATE. 
```
> devtools::install_github("mollyow/estimatr")
> library(estimatr)
> lm_lin(y ~ z_bin-1, covariates = ~ x, data = dat)
            Estimate Std. Error   t value     Pr(>|t|)  CI Lower  CI Upper DF
z_bin0     2.3921681  0.1136183 21.054425 7.934125e-22 2.1617395 2.6225967 36
z_bin1     2.5165919  0.2647700  9.504822 2.370080e-11 1.9796134 3.0535704 36
z_bin0:x_c 0.7151286  0.1249358  5.723967 1.625271e-06 0.4617470 0.9685102 36
z_bin1:x_c 1.1680717  0.3484703  3.351998 1.896537e-03 0.4613412 1.8748022 36
```

A few comments:
 - **(Edit: resolved)** If the new data has additional factor levels/treatment values, prediction will fail as the model matrix will not be the correct dimension. If there are new treatment levels AND old treatment levels are dropped, there could be some bad behavior, as names of treatment levels are not checked. 
 - My understanding is that expected behavior for `predict.lm_robust()` without new data is failure, because the model object does not save the original model matrix. This PR does not modify that behavior, and so doesn't address Luke's question in #403.


tests/testthat/test-lm-robust_margins.R
```diff
@@ -97,17 +97,18 @@ test_that("lm robust + cluster can work with margins", {
 test_that("lm lin can work with margins", {
   skip_if_not_installed("margins")
   data("alo_star_men")
-  lml <- lm_lin(GPA_year1 ~ ssp, ~  gpa0, data = alo_star_men, se_type = "classical")
+  # instruct margins to treat treatment as a factor
+  lml <- lm_lin(GPA_year1 ~ factor(ssp), ~  gpa0, data = alo_star_men, se_type = "classical")
```
mollyow, line 101: My understanding is that margins needs to be instructed which variables are factors, to treat them accordingly. https://cran.r-project.org/web/packages/margins/vignettes/Introduction.html#Using_the_at_Argument

tests/testthat/test-lm-lin.R
```diff
@@ -308,10 +308,19 @@ test_that("Test LM Lin", {
   lmlo <- lm_lin(Y ~ z + 0, ~X1, data = dat)
   expect_equal(
     lmlo$term,
-    c("z", "X1_c", "z:X1_c")
+    c("z0", "z1", "z0:X1_c", "z1:X1_c")
```
mollyow, line 311: The PR changes the expected behavior for binary treatment with no intercept.

R/estimatr_lm_lin.R
```diff
@@ -360,6 +350,12 @@ lm_lin <- function(formula,
 
   return_list[["scaled_center"]] <- center
   setNames(return_list[["scaled_center"]], original_covar_names)
+  # Store unique treatment values
+  if(attr(terms(model_data), "dataClasses")[attr(terms(model_data),"term.labels")[1]] == "factor"){
+    return_list[["treatment_levels"]] <- model_data$xlevels[[1]]
+  } else {
+    return_list[["treatment_levels"]] <- sort(unique(design_matrix[, design_mat_treatment]))
+  }
```
mollyow, line 358: This is added so that when the model matrix is generated for predictions, we can ensure that the new data only includes a subset of treatment levels that were in the original model fit. Without being able to check this, weird behavior could result from predictions where the new data does not share identical treatment levels with the original data.

R/S3_predict.R
```diff
@@ -74,30 +74,6 @@ predict.lm_robust <- function(object,
 
   X <- get_X(object, newdata, na.action)
 
-  # lm_lin scaling
```
mollyow, line 77: all of lm_lin scaling is moved down to `get_X()`

R/S3_predict.R
```diff
@@ -74,30 +74,6 @@ predict.lm_robust <- function(object,
...
-          names(object$scaled_center),
-          drop = FALSE
-        ],
-        center = object$scaled_center,
-        scale = FALSE
-      )
-
-    # Interacted with treatment
-    treat_name <- attr(object$terms, "term.labels")[1]
-    interacted_covars <- X[, treat_name] * demeaned_covars
```
mollyow, line 92: This does not have the desired behavior when there are multiple treatment levels.

R/estimatr_lm_lin.R
```diff
@@ -230,7 +230,7 @@ lm_lin <- function(formula,
   design_mat_treatment <- colnames(design_matrix)[treat_col]
 
   # Check case where treatment is not factor and is not binary
-  if (any(!(treatment %in% c(0, 1)))) {
+  if (any(!(treatment %in% c(0, 1))) | (!has_intercept&ncol(treatment) ==1) ) {
```
mollyow, line 233: This change and the subsequent one modify how `lm_lin()` fits without an intercept and with 0/1 treatment values.

R/estimatr_lm_lin.R
```diff
@@ -313,20 +313,10 @@ lm_lin <- function(formula,
...
     )
   } else {
-    # If no intercept, but treatment is only one column,
-    # need to add base terms for covariates
-    if (n_treat_cols == 1) {
-      X <- cbind(
-        treatment,
-        demeaned_covars,
-        interacted_covars
-      )
```
mollyow, line 323: This special case is resolved

